PR #15, opened 2022-07-21
GitHub actions

amantoux:


Amir-P (commented):
Why are we using a single job called build including analyzing and checking format? Isn't it better to have different job for each one? @amantoux

amantoux:
> Why are we using a single job called build including analyzing and checking format? Isn't it better to have different job for each one? @amantoux 

The install of Java & Flutter is quite long (circa 50s)
(Only reason on my end)

Amir-P (commented):
I think we can get rid of Java since we're not building for Android. Also since we're not building anything it would be more appropriate to use another name instead of build for the job. @amantoux

Amir-P (approved):
LGTM.

PR #16, opened 2022-07-23
Added support for inline embeds

Amir-P:
There are no built-in inline embeds yet but it's possible to define your own embeds. Here is an example (notice the icons):


packages/parchment/lib/src/document/embeds.dart
```diff
@@ -93,14 +89,9 @@ class SpanEmbed extends EmbeddableObject {
 ///
 /// Examples of block embeds include horizontal rule, an image or a map view.
 ///
-/// There are two built-en embed types supported by Parchment documents, however
+/// There are two built-in embed types supported by Parchment documents, however
 /// the document model itself does not make any assumptions about the types
 /// of embedded objects and allows users to define their own types.
-///
```
amantoux, line 99: Is there a reason for removing this comment?

packages/parchment/lib/src/heuristics/delete_rules.dart
```diff
@@ -100,10 +101,17 @@ class PreserveLineStyleOnMergeRule extends DeleteRule {
   }
 }
 
-/// Prevents user from merging a line containing an embed with other lines.
+/// Prevents user from merging a line containing a block embed with other lines.
 class EnsureEmbedLineRule extends DeleteRule {
   const EnsureEmbedLineRule();
 
+  bool isBlockEmbed(Object data) {
```
amantoux, line 108: Can't find where this is used?

amantoux (changes requested):
Two tiny comments.
Also @Amir-P, how can a user extend the tool bar to select an item to be passes to the embed builder?
We keep it for a later PR?

packages/parchment/lib/src/document/embeds.dart
Amir-P: Since it's not the case anymore. Fleather doesn't suppose anything about horizontal rule or image. You can redefine them easily. Am I wrong? @amantoux

packages/parchment/lib/src/heuristics/delete_rules.dart
Amir-P: My bad. Check it again.

packages/parchment/lib/src/document/embeds.dart
amantoux: `FleatherEditor` still implements a defaultEmbedBuilder that supports 'hr", doesn't it?

Amir-P: It doesn't implement it. It uses a default `EmbedBuilder` which is simply overridable as a constructor parameter. @amantoux

amantoux (approved):
✅

PR #20, opened 2022-08-07
Add editorKey to FleatherEditor and FleatherField

Amir-P:
This PR adds `editorKey` to `FleatherEditor` and `FleatherField`'s constructor parameters which is then used as `RawEditor`'s key.

Objective:
This change opens up a lot of possibility for developers to develop third-party packages based on Fleather.
But for now I have two things in mind:
1. My [old mention PR](https://github.com/memspace/zefyr/pull/594) on Zefyr can be turned into a standalone package to cooperate with Fleather and display popup in the proper position using `renderEditor` from `EditorState`.
2. We can develop Medium like editing toolbar as a standalone package again using `renderEditor` from `EditorState`.

### DIFF
--- a/packages/fleather/lib/src/widgets/editor.dart
+++ b/packages/fleather/lib/src/widgets/editor.dart
@@ -202,9 +202,12 @@ class FleatherEditor extends StatefulWidget {
   /// Material [ListTile]s.
   final LinkActionPickerDelegate linkActionPickerDelegate;
 
+  final GlobalKey<EditorState>? editorKey;
+
   const FleatherEditor({
     Key? key,
     required this.controller,
+    this.editorKey,
     this.focusNode,
     this.scrollController,
     this.scrollable = true,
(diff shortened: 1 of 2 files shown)


packages/fleather/lib/src/widgets/editor.dart
```diff
@@ -231,10 +234,10 @@ class FleatherEditor extends StatefulWidget {
```
amantoux, line 234: [nitpick] I see you change to `State<FleatherFIeld>` the return type in `FleatherField`
Can we align & change to `State<FleatherEditor>`

amantoux (changes requested):
Hey @Amit-P
Left a question and a nitpick

packages/fleather/lib/src/widgets/editor.dart
```diff
@@ -246,13 +249,24 @@ class _FleatherEditorState extends State<FleatherEditor>
   late EditorTextSelectionGestureDetectorBuilder
       _selectionGestureDetectorBuilder;
 
-  void _requestKeyboard() {
-    _editorKey.currentState?.requestKeyboard();
+  void _requestKeyboard() => editableTextKey.currentState?.requestKeyboard();
+
+  @override
+  void didUpdateWidget(covariant FleatherEditor oldWidget) {
```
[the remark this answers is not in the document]
Amir-P: Suppose that the `oldWidget.editorKey` is null and `widget.editorKey` is null too. If we do `_editorKey = widget.editorKey ?? GlobalKey<EditorState>()` we will create a new key and lose the old state of editor which is not desired. @amantoux

packages/fleather/lib/src/widgets/editor.dart
Amir-P: Done. @amantoux

amantoux (approved):

PR #21, opened 2022-08-14
[Parchment] Support for background color

amantoux:
This is the first part of a couple of PRs (the second one is related to Fleather)

Simply adds a background color attribute to Parchment.


Amir-P (approved):
LGTM.

PR #24, opened 2022-08-17
Toolbar appears consistently on double-tap

amantoux:
Better show than explain

### Before
toolbar_nok.webm

### After

toolbar_ok.webm


Amir-P (approved):
I haven't had a chance to look into text selection and related things but I think it's ok and based on my tests it's working better than before. So LGTM. @amantoux

PR #27, opened 2022-08-23
Collaborative edition example

cgestes:
This is an example of a collaborative editor for Fleather.

Should we extract the logic somehow to make it reusable?

Note that there is a bug that make the focus switch between the two editors in the example.

Feedback welcome :)

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,25 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "name": "Fleather Example (debug)",
+      "program": "packages/fleather/example/lib/main.dart",
+      "request": "launch",
+      "type": "dart",
+    },
+    {
+      "name": "Fleather Example (profile)",
+      "program": "packages/fleather/example/lib/main.dart",
+      "request": "launch",
+      "type": "dart",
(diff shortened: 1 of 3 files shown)


packages/fleather/example/lib/src/home.dart
```diff
@@ -95,7 +96,7 @@ class _HomePageState extends State<HomePage> {
           color: Colors.grey.shade800,
           child: _buildMenuBar(context),
         ),
-        body: _buildWelcomeEditor(context),
+        body: CollaborativeEditionDemo(), //_buildWelcomeEditor(context),
```
cgestes, line 99: needs to be reverted once finished.

Amir-P:
Thanks for contributing @cgestes . It's good to see you here!
I'm a bit busy these days but I'll try to check it in the next week.

cgestes:
No worries. take your time.

There are a few commits that if they make sense can go in separated PR.

most notably the last one about stealing focus, I tried on Linux and Android and it doesnt seem to have any bad impact.

cgestes:
replaced the demo in this PR by a repository: https://github.com/jotshq/fleather_demo_collab

PR #28, opened 2022-08-25
Cannot make toolbar appear on empty document (immediately after gaining focus)

### LINKED ISSUE #29: Toolbar doesn't appear after a long press while gaining focus
Steps to reproduce

1. Empty document in example (select all - delete)
2. Open example menu, then close it
3. long press on document -> nothing appears (toolbar should appear)

Workaround is to type some text and then the long press (or the double tap) will cause the toolbar to appear.

This is particularly annoying for users who wish to paste something in a new document.
Their impression is that they cannot paste things. They end up not using the product

Video

amantoux:
Fixes #29 

-------

Before fix (go to menu once document is open, emulates opening an empty document and setting focus)

https://user-images.githubusercontent.com/17318211/186650849-74b992b9-dea5-419a-bbe9-e3778ada3ddc.mov

After fix

https://user-images.githubusercontent.com/17318211/186650873-f787b8a9-3a67-4c4d-b504-e238fc7f669d.mov

### DIFF
--- a/packages/fleather/lib/src/rendering/editor.dart
+++ b/packages/fleather/lib/src/rendering/editor.dart
@@ -240,6 +240,7 @@ class RenderEditor extends RenderEditableContainerBox
   }
 
   double? _maxContentWidth;
+
   set maxContentWidth(double? value) {
     if (_maxContentWidth == value) return;
     _maxContentWidth = value;
@@ -606,10 +607,7 @@ class RenderEditor extends RenderEditableContainerBox
     // components, so always send those events, even if we didn't think it
     // changed. Also, the user long pressing should always send a selection change
     // as well.
-    if (selectionChanged ||
-        (cause == SelectionChangedCause.longPress ||
-            cause == SelectionChangedCause.keyboard ||
-            cause == SelectionChangedCause.doubleTap)) {
+    if (selectionChanged || cause.forcesSelectionChanged) {
       onSelectionChanged?.call(nextSelection, cause);
     }
   }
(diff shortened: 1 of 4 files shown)


Amir-P (approved):
LGTM!

Amir-P:
Just a question: Why did you re-add removed TODO? @amantoux

amantoux:
> Just a question: Why did you re-add removed TODOs? @amantoux

Did I? last commit removes them I believe

amantoux:
I've added the same behavior with double tap

_NB: We need to review the whole Editor renderer - widget in order to stick to Flutter latest Editable - EditableText arch; particularly with regards to focus management and user interactions_

packages/fleather/lib/src/widgets/editor.dart
```diff
@@ -26,6 +26,9 @@ import 'text_line.dart';
 import 'text_selection.dart';
 import 'theme.dart';
 
+// TODO: add test - paste on empty document (should not crash on assert)
```
Amir-P: Here you've added them again. @amantoux

packages/fleather/lib/src/rendering/editor.dart
```diff
@@ -608,7 +608,8 @@ class RenderEditor extends RenderEditableContainerBox
     // as well.
     if (selectionChanged ||
         (cause == SelectionChangedCause.longPress ||
-            cause == SelectionChangedCause.keyboard)) {
```
Amir-P: I think it's better to extract this into an extension method on `SelectionChangedCause.longPress`. WDYT? @amantoux

amantoux: @Amir-P what do you mean?

Amir-P: Something like this:
```
Extension on SelectionChangedCause {
  bool get shouldSendEvent => this == SelectionChangedCause.longPress || ...
}
```
WDYT? @amantoux

amantoux: Ok, I'm not sure it will be needed though it the future

Issue #31, opened 2022-09-01
Typing in editor adds the previous typed text after recreating controller


Amir-P:
I've only experienced this issue on Android. Only way to prevent it for now is to recreate editor widget state every time controller changes.

## Steps to Reproduce

1. Type something in editor
2. Recreate controller
3. Type something in editor 

I haven't done any investigations yet but I think this issue roots in our `RawEditorStateTextInputClientMixin` and its `_lastKnownRemoteTextEditingValue`. I believe this issue will be solved after upgrading package to support Flutter 3.3 (By using granular text updates)

Recreating a new controller and passing it to editor I'm getting this log:
```
E/MethodChannel#flutter/textinput(10622): Failed to handle method call
E/MethodChannel#flutter/textinput(10622): java.lang.IndexOutOfBoundsException: invalid composing start: 0
E/MethodChannel#flutter/textinput(10622): 	at io.flutter.embedding.engine.systemchannels.TextInputChannel$TextEditState.<init>(TextInputChannel.java:777)
E/MethodChannel#flutter/textinput(10622): 	at io.flutter.embedding.engine.systemchannels.TextInputChannel$TextEditState.fromJson(TextInputChannel.java:738)
E/MethodChannel#flutter/textinput(10622): 	at io.flutter.embedding.engine.systemchannels.TextInputChannel$1.onMethodCall(TextInputChannel.java:101)
E/MethodChannel#flutter/textinput(10622): 	at io.flutter.plugin.common.MethodChannel$IncomingMethodCallHandler.onMessage(MethodChannel.java:262)
E/MethodChannel#flutter/textinput(10622): 	at io.flutter.embedding.engine.dart.DartMessenger.invokeHandler(DartMessenger.java:295)
E/MethodChannel#flutter/textinput(10622): 	at io.flutter.embedding.engine.dart.DartMessenger.lambda$dispatchMessageToQueue$0$io-flutter-embedding-engine-dart-DartMessenger(DartMessenger.java:319)
E/MethodChannel#flutter/textinput(10622): 	at io.flutter.embedding.engine.dart.DartMessenger$$ExternalSyntheticLambda0.run(Unknown Source:12)
E/MethodChannel#flutter/textinput(10622): 	at android.os.Handler.handleCallback(Handler.java:938)
E/MethodChannel#flutter/textinput(10622): 	at android.os.Handler.dispatchMessage(Handler.java:99)
[...]

PR #35, opened 2022-09-01
Updated RawEditorStateTextInputClientMixin to implement DeltaTextInputClient

### LINKED ISSUE #34: Support Flutter 3.3
Flutter 3.3 again has changes in text editing API and now provides granular text changes (deltas).

Amir-P:
Fixes #34 and #31 .

Thanks to @cgestes for testing and finding selection issue.

### DIFF
--- a/packages/fleather/lib/src/widgets/editor_input_client_mixin.dart
+++ b/packages/fleather/lib/src/widgets/editor_input_client_mixin.dart
@@ -120,6 +120,8 @@ mixin RawEditorStateTextInputClientMixin on EditorState
       return;
     }
 
+    var editingValue = _lastKnownRemoteTextEditingValue!;
+
     for (final textEditingDelta in textEditingDeltas) {
       final delta = Delta();
       if (textEditingDelta is TextEditingDeltaInsertion) {
@@ -127,16 +129,17 @@ mixin RawEditorStateTextInputClientMixin on EditorState
         delta.insert(textEditingDelta.textInserted);
       } else if (textEditingDelta is TextEditingDeltaDeletion) {
         delta.retain(textEditingDelta.deletedRange.start);
-        delta.delete(textEditingDelta.deletedRange.end -
-            textEditingDelta.deletedRange.start);
+        delta.delete(textEditingDelta.deletedRange.length);
       } else if (textEditingDelta is TextEditingDeltaReplacement) {
         delta.retain(textEditingDelta.replacedRange.start);
         delta.insert(textEditingDelta.replacementText);
-        delta.delete(textEditingDelta.replacedRange.end -
-            textEditingDelta.replacedRange.start);
+        delta.delete(textEditingDelta.replacedRange.length);
       }
       widget.controller.compose(delta, selection: textEditingDelta.selection);
+      editingValue = textEditingDelta.apply(editingValue);
     }
+
+    _lastKnownRemoteTextEditingValue = editingValue;
   }
 
   @override
@@ -294,4 +297,8 @@ mixin RawEditorStateTextInputClientMixin on EditorState
       });
     }
   }
+}
+
+extension on TextRange {
+  int get length => end - start;
 }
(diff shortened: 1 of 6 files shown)


Amir-P:
It's still draft but can you test and provide feedback on it? @amantoux

packages/fleather/lib/src/widgets/editor_input_client_mixin.dart
```diff
@@ -120,51 +114,38 @@ mixin RawEditorStateTextInputClientMixin on EditorState
...
+          data = textEditingDelta.textInserted;
+        } else if (textEditingDelta is TextEditingDeltaDeletion) {
+          start = textEditingDelta.deletedRange.start;
+          length = textEditingDelta.deletedRange.length;
+        } else if (textEditingDelta is TextEditingDeltaReplacement) {
+          start = textEditingDelta.replacedRange.start;
+          length = textEditingDelta.replacedRange.length;
+          data = textEditingDelta.replacementText;
+        }
+        widget.controller.replaceText(start, length, data);
```
amantoux, line 141: I think you need to add the new selection coming form the delta when updating the controller
```dart
widget.controller.replaceText(start, length, data, selection: textEditingDelta.selection);
```
This seems to work but might need some more tests

There seems to be a bug with the `suppr` case, the `deletedRange` is the same as a `del` when it should be `TextRange(start:deletedRange.start + 1, end: deletedRange.end + 1)`

amantoux (changes requested):
@amir-P Left a suggestion

packages/fleather/lib/src/widgets/editor_input_client_mixin.dart
Amir-P: I couldn't understand what do you mean by `suppr` case? Also I didn't face any issue with deletion. @amantoux

amantoux: When you hit `suppr` instead of `del` on keyboard, the behavior is incorrect in my tests.
But you have to run it on web (just tried on macOS and there is no issue there)

cgestes:
what's the status of this PR? (have been using daily, works good :))

amantoux:
> what's the status of this PR? (have been using daily, works good :))
There's a problem in Flutter web
Hitting `suppr`, the `deletedRange` is the same as a `del` when it should be `TextRange(start:deletedRange.start + 1, end: deletedRange.end + 1)`

cgestes:
didn't try on the web !

amantoux:
https://github.com/flutter/samples/issues/1424

cgestes:
thanks to @amantoux for pointing out the bugs in flutter/samples.

they made a PR to fix the web issue. https://github.com/flutter/samples/pull/1458

time to go forward?

amantoux:
Don't we need https://github.com/flutter/samples/issues/1424?

Beside, unfortunately (or fortunately depends on the point of view) I'm using this in production, so need to wait for it to land in a release

Amir-P:
Seems like the fix is landed on latest stable release and it handles `suppr` correctly on Web. Can you test too? @cgestes @amantoux 

```Flutter (Channel stable, 3.3.4, on macOS 12.6 21G115 darwin-arm, locale en-US)```

Amir-P:
Looks like the issue with deltas for Web is fixed on 3.7 stable. @amantoux @cgestes

amantoux:
@Amir-P we'll upgrade to Flutter 3.7 with https://github.com/fleather-editor/fleather/pull/89
Can you rebase once this is done?
btw this PR is still draft, is it still intended?

Amir-P:
There were problems with history tests using old text editing API. I've updated them and now I think we're good to go. Please review. @amantoux

amantoux (commented):
@Amir-P, overall it look good to me
I just have concerns about the dependency on Mocktail package (even if only a dev dependency)
It doesn't seem to be maintained (no code update for the last year almost)

PR #37, opened 2022-09-05
Fix new line after inline embed insertion

Amir-P:
Current `InsertEmbedsRule` adds a new line after insertion delta if the data is embeddable object but it should only be applied for block embeds since it's just ensuring that embed is on a new line without any other data before and after it.


amantoux (approved):
LGTM

PR #38, opened 2022-09-06
Review checkbox design

amantoux:
Increase size of checkbox as they are difficult to handle on small devices and review design

### DIFF
--- a/packages/fleather/lib/src/widgets/editable_text_block.dart
+++ b/packages/fleather/lib/src/widgets/editable_text_block.dart
@@ -343,8 +343,6 @@ class _BulletPoint extends StatelessWidget {
   }
 }
 
-const _checkboxSize = 16.0;
-
 class _CheckboxPoint extends StatefulWidget {
   final double width;
   final bool value;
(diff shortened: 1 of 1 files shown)


amantoux:
@Amir-P can you take a look?

amantoux:
@Amir-P simply changed the size of the check box. Tell me if it suits better
Will work on redesign in parallel

Amir-P:
> @Amir-P simply changed the size of the check box. Tell me if it suits better
> Will work on redesign in parallel

Can we use the `Checkbox` from Flutter widgets instead of this custom widget? @amantoux

Amir-P:
It's still small and I guess 24 would be better on touch devices. @amantoux  
https://material.io/components/checkboxes#specs

Amir-P (changes requested):
It's still small and I guess 24 would be better on touch devices. @amantoux  
https://material.io/components/checkboxes#specs

amantoux:
24 is huge, or else implies the line height at least 24 which sounds too much

Try look at this comment in Github mobile app

- [ ] test

Amir-P:
> 24 is huge, or else implies the line height at least 24 which sounds too much @amantoux
> 
> Try look at this comment in Github mobile app
> 
> * [ ]  test

You're right. Let's go with 16. @amantoux

amantoux:
Ok finally no possible customization possible in terms of size on `Checkbox`

Amir-P (approved):
The size is good I think and no need for customization. LGTM.

PR #40, opened 2022-09-08
Fix optional FocusNode causing _CastError

Amir-P:
Although providing `focusNode` is optional for both `FleatherEditor` and `FleatherField` but passing null causes `_CastError`.

### DIFF
--- a/packages/fleather/lib/src/widgets/editor.dart
+++ b/packages/fleather/lib/src/widgets/editor.dart
@@ -765,12 +765,13 @@ class RawEditorState extends EditorState
   FocusNode? _focusNode;
 
   @override
-  FocusNode get effectiveFocusNode => widget.focusNode ?? _focusNode!;
+  FocusNode get effectiveFocusNode =>
+      widget.focusNode ?? (_focusNode ??= FocusNode());
 
   bool get _hasFocus => effectiveFocusNode.hasFocus;
 
   @override
-  bool get wantKeepAlive => effectiveFocusNode.hasFocus;
+  bool get wantKeepAlive => _hasFocus;
 
   TextDirection get _textDirection {
     final result = Directionality.maybeOf(context);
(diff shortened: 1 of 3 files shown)


packages/fleather/lib/src/widgets/editor.dart
```diff
@@ -1013,10 +1006,12 @@ class RawEditorState extends EditorState
...
     if (widget.focusNode != oldWidget.focusNode) {
       oldWidget.focusNode?.removeListener(_handleFocusChanged);
-      _focusNode?.removeListener(_handleFocusChanged);
-      _focusNode?.dispose();
-      _focusNode = null;
-      _createInternalFocusNodeIfNeeded();
+      if (widget.focusNode != null) {
+        _focusNode
+          ?..removeListener(_handleFocusChanged)
+          ..dispose();
```
cgestes, line 1012: maybe add a comment somewhere to explain _focusNode is a potential internal focusNode? (or name it _internalFocusNode?)

Just to make it clear that it has to be handled only when we created it because a user one was not provided.

Amir-P: Done. @cgestes

amantoux (commented):
@Amit-P, PTAL

amantoux:
While implementing test, cursorStyle caused null value problems...
It is never call with a null value in `RawEditor` so I've put it as `required` In the constructor
Let me know if any comment

Amir-P:
> While implementing test, cursorStyle caused null value problems... It is never called with a `null` value in `RawEditor` so I've put it as `required` In the constructor Besides, we should ensure test coverage of `RawEditor::didUpdateWidget` generally speaking
> 
> Let me know your thoughts

Tests are better know with the changes you've made. But we're still missing tests for `_FleatherFieldState::didUpdateWidget`. @amantoux

PR #41, opened 2022-09-08
Small fixes

cgestes:
- Add a vscode launch configuration to easily launch the example
- stop throwing when _textInputConnection is null (this happens on Linux from time to time)
- stop requesting focus/keyboard every time the text changes, this makes it impossible to have two collaborative editor on the same page. (one editor steal the focus of the other when receiving changes otherwise)

The last change probably needs serious testing on all platform. (there must be a reason it was there..)

I can rebase, split into smaller changes if needed.


Amir-P:
I'm ok with all of the changes but I think the request for keyboard on document update is intentional (in case user enables or disables any attribute from toolbar) and we should add it to the toolbar instead of requesting it in every change of document for cases like collaborative editing.
WDYT? @amantoux @cgestes

Amir-P (commented):
I'm ok with all of the changes but I think the request for keyboard on document update is intentional (in case user enables or disables any attribute from toolbar) and we should add it to the toolbar instead of requesting it in every change of document for cases like collaborative editing.
WDYT? @amantoux @cgestes

cgestes:
Make sense, I'll give it a shot.

amantoux (commented):
Fine by me

Amir-P:
Then let's merge this and fix toolbar focus request in another branch. 
@cgestes Can you please format using dart? Our pipeline is complaining about it.

amantoux:
> Then let's merge this and fix toolbar focus request in another branch.
> @cgestes Can you please format using dart? Our pipeline is complaining about it.

You're sure you don't want to fix before merging into master? Doesn't this make the whole thing unusable?

Amir-P:
> > Then let's merge this and fix toolbar focus request in another branch.
> > @cgestes Can you please format using dart? Our pipeline is complaining about it.
> 
> You're sure you don't want to fix before merging into master? Doesn't this make the whole thing unusable?

We're not releasing it without fixing that but we can do it first too. @cgestes

cgestes:
Yep. Will format and send the others changes and eventually fix the real bug properly

PR #42, opened 2022-09-18
Release v1.2.2

Amir-P:
Removed dependency override of local Parchment from example since it depends on local Fleather and local Fleather depends on local Parchment. So no need for that anymore.


amantoux (approved):
lgtm

PR #43, opened 2022-09-26
Support document history

amantoux:
Support undoing & redoing a document's history using default shortcuts of the platform (`ctrl+z`, `ctrl+shift+z`, etc..)

### DIFF
--- a/.github/workflows/fleather.yml
+++ b/.github/workflows/fleather.yml
@@ -10,7 +10,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: subosito/flutter-action@v2
         with:
-          flutter-version: "3.0.1"
+          flutter-version: "3.3.2"
           cache: true
           # Manually Update this `key`
           cache-key: "20220721"
@@ -35,7 +35,7 @@ jobs:
 
       - name: Parchment - Run tests
         working-directory: ./packages/parchment
-        run: | 
+        run: |
           dart test --coverage coverage
           dart pub global run coverage:format_coverage --lcov --in ./coverage --out ./coverage/lcov.info --report-on:lib
 
@@ -59,11 +59,9 @@ jobs:
       - name: Fleather - Run tests
         working-directory: ./packages/fleather
         run: flutter test --coverage
-        
+
       - name: Codecov
         uses: codecov/codecov-action@v3.1.0
         with:
           files: ./packages/fleather/coverage/lcov.info,./packages/parchment/coverage/lcov.info
           fail_ci_if_error: true
-
-
--- a/packages/fleather/lib/src/widgets/history.dart
+++ b/packages/fleather/lib/src/widgets/history.dart
@@ -33,8 +33,8 @@ class FleatherHistory extends StatefulWidget {
 }
 
 class _FleatherHistoryState extends State<FleatherHistory> {
-  late final HistoryStack _stack;
   late final _Throttled<Delta> _throttledPush;
+  late HistoryStack _stack;
   Timer? _throttleTimer;
 
   // This duration was chosen as a best fit for the behavior of Mac, Linux,
--- a/packages/fleather/lib/src/widgets/history.dart
+++ b/packages/fleather/lib/src/widgets/history.dart
@@ -1,0 +1,317 @@
+import 'dart:async';
+
+import 'package:fleather/fleather.dart';
+import 'package:flutter/widgets.dart';
+import 'package:quill_delta/quill_delta.dart';
+
+/// A void function that takes a [TextEditingValue].
+@visibleForTesting
+typedef TextEditingValueCallback = void Function(TextEditingValue value);
+
+/// Provides undo/redo capabilities for text editing.
+///
+/// Listens to [controller] as a [ValueNotifier] and saves relevant values for
+/// undoing/redoing. The cadence at which values are saved is a best
+/// approximation of the native behaviors of a hardware keyboard on Flutter's
+/// desktop platforms, as there are subtle differences between each of these
+/// platforms.
+///
+/// Listens to keyboard undo/redo shortcuts and send update to [controller].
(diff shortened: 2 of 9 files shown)


amantoux:
@Amir-P can you please take a look and test it on your end?

amantoux:
@cgestes do you want to take a look?

.github/workflows/fleather.yml
```diff
@@ -10,7 +10,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: subosito/flutter-action@v2
         with:
-          flutter-version: "3.0.1"
+          flutter-version: "3.3.2"
           cache: true
           # Manually Update this `key`
           cache-key: "20220721"
```
cgestes, line 16: should this be updated at the same time?

packages/fleather/lib/src/widgets/history.dart
```diff
@@ -0,0 +1,280 @@
...
+      extentOffset = changeDelta.textLength;
+    }
+    if (lastOp.isInsert) {
+      baseOffset = changeDelta.textLength;
+      extentOffset = baseOffset;
+    }
+    return TextSelection(baseOffset: baseOffset, extentOffset: extentOffset);
+  }
+
+  void _push() {
```
cgestes, line 81: rename to _onReceiveLocalChanges?

packages/fleather/lib/src/widgets/history.dart
```diff
@@ -0,0 +1,280 @@
...
+    if (widget.controller.plainTextEditingValue == TextEditingValue.empty) {
+      return;
+    }
+    _throttleTimer = _throttledPush(widget.controller.document.toDelta());
```
cgestes, line 85: why not using changes instead of diffing full documents?

(and aggregate them to the last element of the stack if needed)

packages/fleather/lib/src/widgets/history.dart
```diff
@@ -0,0 +1,280 @@
...
+    widget.controller.compose(changeDelta,
+        selection: _fromDelta(changeDelta), source: ChangeSource.local);
+  }
+
+  TextSelection _fromDelta(Delta changeDelta) {
+    assert(changeDelta.isNotEmpty);
+    final firstOp = changeDelta.first;
+    int baseOffset = 0;
+    if (firstOp.isRetain && firstOp.attributes == null) {
+      baseOffset = firstOp.length;
```
cgestes, line 67: add a comment?
I would explain that if the operation is a retain without attributes it is just the operations to go to the change, it is not the actual change.

packages/fleather/lib/src/widgets/history.dart
```diff
@@ -0,0 +1,280 @@
...
+    }
+    int extentOffset = baseOffset;
+    final lastOp = changeDelta.last;
+    if (lastOp.isRetain && lastOp.attributes != null) {
+      extentOffset = changeDelta.textLength;
+    }
+    if (lastOp.isInsert) {
+      baseOffset = changeDelta.textLength;
```
cgestes, line 75: add a comment?
explain why we would like the cursor to be at the beginning of the selection ?

cgestes (commented):
here we go :)

Happy to see that feature brewing :)

packages/fleather/lib/src/widgets/history.dart
amantoux: would love to 
Any idea where to source those changes?
I was thinking we could enhance controller once we move to leveraging on `DeltaTextEditingValue`

cgestes: ```controller.document.changes```

amantoux: Ok but then you will capture all changes even those coming from a remote source

amantoux: I'll give it a try (sorry this comes after a long (much interrupted) thought)

cgestes: remote or not you need to capture them, you probably only want to undo your local change, but you will need to rebase to transpose them over the remote changes.

amantoux: Yes thought so too

amantoux: I'm starting to implement but several questions arise:

1. Rebasing is not that simple. I enter `hello world` then I format in **bold** .
   Someone overwrites all with `bye`
   
   **What is the expected behavior?**
2. **If** we want to mimic the thottling that Flutter has implemented (which I believe we should but shout if you think otherwise), we need to implement a coalescing mechanism whereby when we compose any sequence of local `ParchmentChange` and any sequence of remote `ParchmentChange` during the throttle timer
   (`{loc, loc loc},{remote,remote},{loc}...`) to treat them at the end of the timer.

   While this is possible, it sounds overcomplicated for a first version

Happy to read your thoughts on this

amantoux: 1. Ok but we need to agree on a algo to handle all cases
2. Not complicated but given the above I think we should leave it aside and handle remote changes in a separate PR
3. I think we should review the source of changes in general not only for the undo/redo, again a separate PR. We might face performance issues on very long documents so I'd adress this first

cgestes: 1. I believe this is already coded properly. (just use Delta.transform)
2. yep
3. how would you implement it without knowing the change is coming from undo? adding some kind of metadata to the change? comparing md5?

amantoux: 1. Yes `transform` is convenient, but we need to decide who takes precedence and when etc... more a specification work

3. Dunno yet. Maybe simply add a `undo_redo` to `hangeSource`. Didn't think this through

PR #45, opened 2022-09-27
Feat/remote selections

cgestes:
What do you think of adding support for displaying multiple remote selections?

What do you think of the direction of this patch?

(I suspect to do it really nicely we also need to add overlays where we display the name associated to each selections).

Probably some fade-in / fade-out also.

### DIFF
--- a/packages/fleather/lib/src/widgets/controller.dart
+++ b/packages/fleather/lib/src/widgets/controller.dart
@@ -14,6 +14,13 @@ List<String> _insertionToggleableStyleKeys = [
   ParchmentAttribute.inlineCode.key,
 ];
 
+class RemoteSelection {
+  final String name;
+  final TextSelection selection;
+  final Color color;
+  RemoteSelection(this.name, this.color, this.selection);
+}
+
 class FleatherController extends ChangeNotifier {
   FleatherController([ParchmentDocument? document])
       : document = document ?? ParchmentDocument(),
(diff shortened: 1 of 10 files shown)


amantoux:
Took a glance, sounds nice

I'll try to find some time to look through it next week

We're focusing on providing basic editing features & stability at the moment

cgestes:
ok no worries. It can wait :)

Amir-P:
Hey @cgestes , Why did you close the PR? I was going to review it these days.

cgestes:
I thought I would republish later when more elaborated and it would be a better timing for the project. this version while functional is quite simple.

packages/fleather/lib/src/widgets/controller.dart
```diff
@@ -26,6 +33,14 @@ class FleatherController extends ChangeNotifier {
   TextSelection get selection => _selection;
   TextSelection _selection;
 
+  List<RemoteSelection> get remoteSelections => _remoteSelections;
```
Amir-P, line 36: Isn't it better to pass `remoteSelections` directly to widget (Editor or Field) instead of controller?

packages/fleather/lib/src/widgets/controller.dart
```diff
@@ -14,6 +14,13 @@ List<String> _insertionToggleableStyleKeys = [
   ParchmentAttribute.inlineCode.key,
 ];
 
+class RemoteSelection {
```
Amir-P, line 17: I think it's enough for now but we can have a builder function to build overlays with cursors too.

Amir-P (commented):
PTAL. @cgestes

packages/fleather/lib/src/widgets/controller.dart
cgestes: yeah this will come.

packages/fleather/lib/src/widgets/controller.dart
Amir-P: I think it's more natural to provide selections in `build` like other widgets. But if you're sure that we're gonna need some optimizations in future I'm ok with the way it's now. @cgestes 
Also what do you think about this? @amantoux

cgestes: ? I'am not sure I am following you.
What changes often seems to often be provided by a controller. I dont know what is better, the whole infrastructure was there to support that usecase easily in fleather. (it's already done for the local selection)

PR #49, opened 2022-10-11
force a refresh when focus changes

cgestes:
otherwise _hasFocus for EditableTextLine is false while the widget has the focus and the cursor is not displayed.

This is triggered when the text field is empty. (not always, but most of the time).

In practice: you click on an empty Fleather editor and nothing happens. (the cursor doesn't start blinking).
I also tried to remove the the _hasFocus check from https://github.com/fleather-editor/fleather/blob/master/packages/fleather/lib/src/rendering/editable_text_line.dart#L731 and it worked perfectly from my test. (not printing the cursor when it was not needed...so if you know what is best...)


Amir-P (approved):
LGTM.

PR #50, opened 2022-10-27
Release v1.3.0

amantoux:


Amir-P (commented):
It's a bit late to say this since it's released on pub.dev but there were no need to release a new version of Parchment since there were no change in it. Also Parchment changelog doesn't include 1.3.0 changes and running `dart pub publish` should've warned you.

amantoux:
> It's a bit late to say this since it's released on pub.dev but there were no need to release a new version of Parchment since there were no change in it. Also Parchment changelog doesn't include 1.3.0 changes and running `dart pub publish` should've warned you.

Yes it's simply that I fancy the idea of aligning versions (not patches) between Fleather & Parchment
I'd even be in favor of bundling them together

WDYT?

Amir-P:
> > It's a bit late to say this since it's released on pub.dev but there were no need to release a new version of Parchment since there were no change in it. Also Parchment changelog doesn't include 1.3.0 changes and running `dart pub publish` should've warned you.
> 
> Yes it's simply that I fancy the idea of aligning versions (not patches) between Fleather & Parchment I'd even be in favor of bundling them together
> 
> WDYT?

1. I'm not a fan of releasing just for aligning versions. Parchment and Fleather are separate packages.
2. By bundling, do you mean combining them into a single plugin? @amantoux

amantoux:
> > > It's a bit late to say this since it's released on pub.dev but there were no need to release a new version of Parchment since there were no change in it. Also Parchment changelog doesn't include 1.3.0 changes and running `dart pub publish` should've warned you.
> > 
> > 
> > Yes it's simply that I fancy the idea of aligning versions (not patches) between Fleather & Parchment I'd even be in favor of bundling them together
> > WDYT?
> 
> 1. I'm not a fan of releasing just for aligning versions. Parchment and Fleather are separate packages.
> 2. By bundling, do you mean combining them into a single plugin? @amantoux

1. yeah but managed as a single code base
2. yes combining them, separation of concerns doesn't necessarily mean separate packages

Amir-P:
> > > > It's a bit late to say this since it's released on pub.dev but there were no need to release a new version of Parchment since there were no change in it. Also Parchment changelog doesn't include 1.3.0 changes and running `dart pub publish` should've warned you.
> > > 
> > > 
> > > Yes it's simply that I fancy the idea of aligning versions (not patches) between Fleather & Parchment I'd even be in favor of bundling them together
> > > WDYT?
> > 
> > 
> > 
> > 1. I'm not a fan of releasing just for aligning versions. Parchment and Fleather are separate packages.
> > 2. By bundling, do you mean combining them into a single plugin? @amantoux
> 
> 1. yeah but managed as a single code base
> 2. yes combining them, separation of concerns doesn't necessarily mean separate packages

I agree with "separation of concerns doesn't necessarily mean separate packages", but Parchment and Fleather being separated isn't because of SoC. Parchment doesn't depend on Flutter and you can use it in Dart projects without the editor and without depending on Flutter.

amantoux:
> > > > > It's a bit late to say this since it's released on pub.dev but there were no need to release a new version of Parchment since there were no change in it. Also Parchment changelog doesn't include 1.3.0 changes and running `dart pub publish` should've warned you.
> > > > 
> > > > 
> > > > Yes it's simply that I fancy the idea of aligning versions (not patches) between Fleather & Parchment I'd even be in favor of bundling them together
> > > > WDYT?
> > > 
> > > 
> > > 
> > > 1. I'm not a fan of releasing just for aligning versions. Parchment and Fleather are separate packages.
> > > 2. By bundling, do you mean combining them into a single plugin? @amantoux
> > 
> > 
> > 
> > 1. yeah but managed as a single code base
> > 2. yes combining them, separation of concerns doesn't necessarily mean separate packages
> 
> I agree with "separation of concerns doesn't necessarily mean separate packages", but Parchment and Fleather being separated isn't because of SoC. Parchment doesn't depend on Flutter and you can use it in Dart projects without the editor and without depending on Flutter.

Maybe the question sums up to whether we intend to propose Parchment as a product of its own.
I personally don't anticipate supporting Parchment for other usages that Fleather. Should we?

Amir-P:
> Maybe the question sums up to whether we intend to propose Parchment as a product of its own. I personally don't anticipate supporting Parchment for other usages that Fleather. Should we?

I'm not using Parchment without Fleather but since it's not a hard thing to do and doesn't add up to the work we have to do to support them separately, I think we should continue to do so.

amantoux:
> > Maybe the question sums up to whether we intend to propose Parchment as a product of its own. I personally don't anticipate supporting Parchment for other usages that Fleather. Should we?
> 
> 
> 
> I'm not using Parchment without Fleather but since it's not a hard thing to do and doesn't add up to the work we have to do to support them separately, I think we should continue to do so.

I disagree 
But never mind

amantoux:
So Release 1.3.0 will have a Parchment 1.2.1 and potentially release 7.0.0 will still have Parchment 1.2.1
We must have a rule of some sort don't you think?

Amir-P:
> We must have a rule of some sort don't you think?

Not necessarily. What is the problem of `Fleather v7.0.0` depending on `Parchment v1.2.1`? `Fleather v1.3.0` depends on `collection v1.16.0` and `Fleather v7.0.0` might depend on collection `v1.17.0`. Is that a problem too?

amantoux:
Of course not; anything is possible from a technical perspective
My point is from a product management perspective: do you want to support Parchment evolutions for another consumer?

If someone files an issue for an evolution of Parchment for another consumer than Fleather, I don't think we will handle it - it's all driven by Fleather.

PR #55, opened 2022-12-05
Support for HTML decoding/encoding

### LINKED ISSUE #56: Rich clipboard
When copy some text (including embeds), pasting it back in the same Fleather widget should preserve format & embeds.

Three use-cases are foreseen:
1. #### Copy/paste from/to a Fleather editor to/from an external app. 
    We assume that the OS clipboard holds HTML or plain text. The content should be converted to Delta thanks to HTML codec. Copying some text from Fleather to the clipboard will convert ParchmentDocument Deltas to HTML prior to setting it to the clipboard.
2. #### Copy from a Fleather editor to another Fleather editor. 
    Copying from one Fleather editor to another Fleather editor should be using Delta format
3. #### Copy from a position to another position within a same Fleather editor
     Same as 2.

amantoux:
This PR is necessary to close #56.
Parchment can now convert a document `Delta` to `HTML` and convert `HTML` to `Delta`.

### Encoding :
- Checklists give the following result:
```html 
<div class="checklist">
  <div class="checklist-item><input type="checkbox" checked><label>item 1</label></div>
  <div class="checklist-item><input type="checkbox"><label>item 2</label></div>
</div>
```
- Indentations on blocks that are not lists give the following
```html
<p style="padding-left:32px">indented text</p>
```
- Directionality is converted as follows:
```html
<p dir="rtl">العربية</p>
```
- Single paragraphs (plain blocks) are encoded to simple text (no `<p>`)
```html
A single <strong>paragraph</strong>
```

### Decoding :
- Checklists are not supported as there are numerous ways to layout `<input>`s
- Indentations in blocks that are not lists are not supported due to the numerous way of materializing in `HTML`
- `<br>`'s are ignored as they do not map to any attribute in Fleather (new line lead necessarily to a new block)

### DIFF
--- a/packages/parchment/lib/src/convert/html.dart
+++ b/packages/parchment/lib/src/convert/html.dart
@@ -1,0 +1,881 @@
+import 'dart:convert';
+
+import 'package:collection/collection.dart';
+import 'package:html/dom.dart' as html;
+import 'package:html/parser.dart';
+import 'package:parchment/parchment.dart';
+import 'package:quill_delta/quill_delta.dart';
+
+extension on ParchmentStyle {
+  Iterable<ParchmentAttribute> get lineAttributes =>
+      values.where((e) => e.scope == ParchmentAttributeScope.line);
+
+  Iterable<ParchmentAttribute> get inlineAttributes =>
+      values.where((e) => e.scope == ParchmentAttributeScope.inline);
+}
+
+final _inlineAttributesParchmentToHtml = {
+  ParchmentAttribute.bold.key: 'strong',
+  ParchmentAttribute.italic.key: 'em',
+  ParchmentAttribute.underline.key: 'u',
+  ParchmentAttribute.strikethrough.key: 'del',
+  ParchmentAttribute.inlineCode.key: 'code',
+  ParchmentAttribute.link.key: 'a',
+};
+
+const _indentWidthInPx = 32;
+
+/// HTML conversion of Parchment
+///
+/// *note: `<br>` are not recognized as new lines and will be ignored*
+///
+/// ## Inline mapping
+/// - b -> <strong>
+/// - i -> <em>
+/// - u -> <u>
+/// - s -> <del>
+/// - c -> <code>
+/// - a -> <a>
+///
+/// ### Line mapping
+/// - default -> <p>
+/// - heading X -> <hX>
+/// - bq -> <blockquote>
+/// - code -> <pre><code>
+/// - ol -> <ol><li>
+/// - ul -> <ul><li>
+/// - cl -> <div class="checklist">
+///           <div class"checklist-item><input type="checklist" checked><label>
+/// - alignment -> <xxx align="left | right | center | justify">
+/// - direction -> <xxx dir="rtl">
+///
+class ParchmentHtmlCodec extends Codec<Delta, String> {
+  const ParchmentHtmlCodec();
+
+  @override
+  Converter<String, Delta> get decoder => const _ParchmentHtmlDecoder();
+
+  @override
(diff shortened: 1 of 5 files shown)


amantoux:
@cgestes @Amir-P if you time to take a look, I'd be glad to have some feedback

packages/parchment/lib/src/convert/html.dart
```diff
@@ -0,0 +1,881 @@
...
+  ParchmentAttribute.strikethrough.key: 'del',
+  ParchmentAttribute.inlineCode.key: 'code',
+  ParchmentAttribute.link.key: 'a',
+};
+
+const _indentWidthInPx = 32;
+
+/// HTML conversion of Parchment
+///
+/// *note: `<br>` are not recognized as new lines and will be ignored*
```
cgestes, line 30: can it be fixed?

dsyrstad:
This looks great! I just spent a day writing an HTML encoder (not the decoder). It would be great if we could merge this!

packages/parchment/lib/src/convert/html.dart
```diff
@@ -0,0 +1,881 @@
...
+    final p = previous.style.values;
+    final c = current.style.values;
+    final areOrderedUnorderedLists = (p.contains(ParchmentAttribute.ol) ||
+            p.contains(ParchmentAttribute.ul)) &&
+        (c.contains(ParchmentAttribute.ol) ||
+            c.contains(ParchmentAttribute.ul));
+    final areChecklists =
+        p.contains(ParchmentAttribute.cl) && c.contains(ParchmentAttribute.cl);
+    if (areOrderedUnorderedLists || areChecklists) {
+      final cssAttributes = [
```
cgestes, line 101: cssAttributes?

packages/parchment/lib/src/convert/html.dart
```diff
@@ -0,0 +1,881 @@
...
+        ParchmentAttribute.alignment.unset,
+        ParchmentAttribute.alignment.center,
+        ParchmentAttribute.alignment.right,
+        ParchmentAttribute.alignment.justify,
+        ParchmentAttribute.direction.rtl,
+        ParchmentAttribute.direction.unset,
+      ];
+      final modifiedPrevious = previous.style.removeAll(cssAttributes);
```
cgestes, line 109: what are we doing here?

packages/parchment/lib/src/convert/html.dart
```diff
@@ -0,0 +1,881 @@
...
+    int nextLineStartPosition = 0;
+
+    for (final op in input.toList()) {
+      if (_hasPlainParagraph(op)) {
+        _processInlineTags(op, buffer, openInlineTags);
+        nextLineStartPosition = _handlePlainBlock(op, buffer, openBlockTags);
+        continue;
+      }
+
+      _processInlineTags(op, buffer, openInlineTags);
```
cgestes, line 165: neatpick: this can be put before the if to make it clear it happens on all iteration of the for no matter what. (and removed from the first if of the for)

amantoux:
Thanks for the code review @cgestes 

I had also (mainly) concerns about how Deltas would map to HTML
I'm not an expert in the field
Any input welcome

Bear in mind that this enables the copy/paste to/from external apps

amantoux:
> This looks great! I just spent a day writing an HTML encoder (not the decoder). It would be great if we could merge this!

Should be shortly

cgestes:
Ok. Not a specialist either. Markdown is not possible for that? Seems more appropriate for copy/paste than html that can express much more than we need. 

Btw I am working on mentions. :)

PR #66, opened 2022-12-10
Support modifying document rules

zaynetro:
This is WIP implementation of https://github.com/fleather-editor/fleather/discussions/62 .

Atm it works without introducing a new embed type. I like the simplicity of it but it might not be the case in real life scenarios (e.g horizontal rule must be the only child on a line). So maybe we can add a constraint on horizontal rule that it must be lone child or let it up to the user?

How does it work?

I have implemented `GroupEmbedProxy` (I don't like the name :/). This new proxy class simply delegates laying out the children to Flutter's `Flex` widget. I had to create new `RenderGroupEmbedProxy` class so that it implements caret methods.

What I need help with:

* Feedback
* Testing
* Pointers where and how to write tests (I am new to Flutter)

What I am planning to work on more:

* Fix: Inserting text between the embeds only splits new line and doesn't show the text. Text later reappears after inserting new line.
* Fix: It is not possible to remove new line between two embeds.

[Screencast from 2022-12-10 14-35-41.webm](https://user-images.githubusercontent.com/627197/206855942-d23fded8-fa91-4387-8b77-ec847b0dee48.webm)


packages/fleather/lib/src/rendering/embed_proxy.dart
```diff
@@ -50,3 +50,93 @@ class RenderEmbedProxy extends RenderProxyBox implements RenderContentProxyBox {
...
+/// render boxes.
+class RenderGroupEmbedProxy extends RenderFlex
+    implements RenderContentProxyBox {
+  RenderGroupEmbedProxy({
+    TextDirection? textDirection,
+  }) : super(textDirection: textDirection);
+
+  @override
+  List<TextBox> getBoxesForSelection(TextSelection selection) {
+    return [];
```
zaynetro, line 66: Need to implement this somehow.

packages/fleather/lib/src/rendering/embed_proxy.dart
```diff
@@ -50,3 +50,93 @@ class RenderEmbedProxy extends RenderProxyBox implements RenderContentProxyBox {
...
+      if (i == 0) {
+        child = firstChild!;
+      } else {
+        child = childAfter(child!)!;
+      }
+
+      // TODO: Can we use `semanticBounds` here?
+      // As an alternative I can also calculate the size myself:
+      //   final width = (child as RenderBox).getMinIntrinsicWidth(size.height);
+      final width = child.semanticBounds.width;
```
zaynetro, line 97: How do I get child's width correctly?

packages/fleather/lib/src/rendering/embed_proxy.dart
```diff
@@ -50,3 +50,93 @@ class RenderEmbedProxy extends RenderProxyBox implements RenderContentProxyBox {
...
+    for (var i = 0; i < childCount; i += 1) {
+      if (i == 0) {
+        child = firstChild!;
+      } else {
+        child = childAfter(child!)!;
+      }
+
+      // I don't know what to use. As an alternative I can also calculate the size myself:
+      //   final width = (child as RenderBox).getMinIntrinsicWidth(size.height);
+      final halfWidth = child.semanticBounds.width / 2;
```
zaynetro, line 122: Same question as above.

cgestes:
I will give it a more thorough review later. 
it seems a bit off to turn block embeds into inline embeds just because they follow each other. Both behaviours seems wanted. (Multiple images per line or single image per line). To get multiple images per line inline embeds seems more appropriate. 

On another hand it probably make sense to have a general way of being able to make advanced layouts in the editor. (Grid of blocs for example). I don’t see reasons for it to be restricted to images though.

zaynetro:
Yes, my idea was that it would be nice to support grouping any type of embed.
And yes, there needs to be a way to limit whether some embeds could be grouped or not (e.g horizonal rule).

At the moment it is not possible to join embed line with any other line, only manually through the controller. So it kind of works as a limiter. Might be a good solution for now. 🤷‍♂️

amantoux:
Why aren't you using `SpanEmbed`?

zaynetro:
> Why aren't you using `SpanEmbed`?

I don't want to inline images with text on a single line. I want to be able to group blocks on a single line.

My use case: users can add multiple images and multiple files. I want to group images together and to group files together.

Here's how it might look:

image

My current understanding is that `SpanEmbed`s are for inlining blocks with text and `BlockEmbed` is for having no text at all on the same line.

amantoux:
I see. This behavior doesn't seem standard though.
You might want implement an ad-hoc `BlockEmbed` that provides this behavior. 
Is there any thing that prevents you from doing what you wish to achieve?

zaynetro:
> You might want implement an ad-hoc `BlockEmbed` that provides this behavior.

Do you mean implement a single embed block that will display a list of images in `FleatherEditor.embedBuilder`?
I can but the problem is that if user puts cursor after my embed and hits backspace then my whole embed will be removed from the document.

Obviously, I can listen to the changes and only apply them partially but I was thinking if we could find a better solution.
Or did you mean a different solution?


> This behavior doesn't seem standard though.

Here is an example of Apple Notes: 

https://user-images.githubusercontent.com/627197/206916461-10f9c690-9958-4c76-b81a-056577b27f95.MP4

What I like is that cursor works just as it would with text. You can separate a list of images and into multiple lines or join images together.

zaynetro:
And there is another example of samle Apple Notes:

https://user-images.githubusercontent.com/627197/206916771-dde5fc35-df0b-47a4-80d5-17b5c39f9c2e.MP4

I really like how images (embeds) become just part of text but have this restriction that no text is allowed on the same line.

amantoux:
Ok, I don't understand why we would want to prevent user's from mixing text & embeds.
MS Word or Google Doc let you do so, which makes me wonder why Apple implemented this limitation.

If it were the case, shouldn't we rather prevent this via an insert rule (if previous node is inline embed and insertion is text, go to line)?

cgestes:
Gave it a try, it is really nice.
Basically if blocks embeds are not followed by a line carriage they are rendered in a flex box? seems pretty intuitive to use :)

(it is still possible to have inline blocks to mix them with text, which seems a less modern way of doing)

Some feedback: I noticed that it fails with selection currently. If we press return between two pictures, it splits the flex into two flexes. (pretty cool). it would be nice if we could have the reverse operation (removing the new line would merge back the two flexes, I suspect this doesnt work because of the selection issues).

packages/fleather/lib/src/rendering/embed_proxy.dart
```diff
@@ -50,3 +50,93 @@ class RenderEmbedProxy extends RenderProxyBox implements RenderContentProxyBox {
...
+    return preferredLineHeight;
+  }
+
+  @override
+  Offset getOffsetForCaret(TextPosition position, Rect caretPrototype) {
+    if (childCount == 0 || position.offset == 0) {
+      return Offset.zero;
+    }
+
+    RenderBox? child;
```
cgestes, line 80: = firstChild

packages/fleather/lib/src/rendering/embed_proxy.dart
```diff
@@ -50,3 +50,93 @@ class RenderEmbedProxy extends RenderProxyBox implements RenderContentProxyBox {
...
+
+    for (var i = 0; i < childCount; i += 1) {
+      if (position.offset <= i) {
+        return Offset(accWidth, 0.0);
+      }
+
+      if (i == 0) {
+        child = firstChild!;
+      } else {
+        child = childAfter(child!)!;
```
cgestes, line 91: we probably can get ride of the whole if here and just update child as the last operation of the for loop.

packages/fleather/lib/src/rendering/embed_proxy.dart
cgestes: can we use something like:
```
child = children[position]; 
return child.localToGlobal(Offset(child.size.with, 0), this);
```
to get the offset of a child directly? (without needed a for loop)

packages/fleather/lib/src/rendering/embed_proxy.dart
cgestes: child.size.width?

cgestes: (see the methods of RenderEditableContainerBox which are pretty similar to what we want to achieve here, and how they are used by RenderEditor for example)

amantoux:
> Ok, I don't understand why we would want to prevent users from mixing text & embeds. MS Word or Google Doc let you do so, which makes me wonder why Apple implemented this limitation.
> 
> If it were the case, shouldn't we rather prevent this via an insert rule (if previous node is inline embed and insertion is text, go to line)?

After a little digging, it seems there are issues with `SpanEmbed` when dealing with heights due to usage of `WidgetSpan`. (see https://github.com/fleather-editor/fleather/issues/67)

zaynetro:
> After a little digging, it seems there are issues with `SpanEmbed` when dealing with heights due to usage of `WidgetSpan`. (see #67)

I saw support for inline images in flutter_quill package. These are the implementation references: 

* https://www.programmersought.com/article/68774767117/
* https://github.com/singerdmx/flutter-quill/issues/323

> I noticed that it fails with selection currently.

I haven't implemented `getBoxesForSelection` method. Probably related.

 > removing the new line would merge back the two flexes
 
 That would be nice to have. I think that's because there are insertion and deletion rules with special cases for embed blocks. Haven't dug into that area yet.
 
 <hr />
 
 I see two paths forward each with its own twists :)
 
 1. Continue this approach with being able to group block embeds on a single line.
     It already kind of works but needs selection to work and insertion/deletion rule changes to allow joining two block lines into one. Also it needs to have a way of dealing with exclusive blocks like horizontal rule.
     
2. Implement truly inline blocks based on the links (I will try to take a look at details later).
    Then we also need some events/hooks as @amantoux suggested for insertion/deletion rules.
    
For insertion/deletion rules it would be nice to ask the developer: "Hey, this text is going to be removed/added, here is the text below and after. What shall we do?" And then developer either continues or aborts the change. As a coincidence these customizable rules would solve both (1) and (2).

amantoux:
> I saw support for inline images in flutter_quill package. These are the implementation references:
>
> https://www.programmersought.com/article/68774767117/
> https://github.com/singerdmx/flutter-quill/issues/323

Yes, solution is the same as implementing `SpanEmbed` (happens I found a way to solve the issue)

> I see two paths forward each with its own twists :)
> 
> Continue this approach with being able to group block embeds on a single line.
> It already kind of works but needs selection to work and insertion/deletion rule changes to allow joining two block lines into one. Also it needs to have a way of dealing with exclusive blocks like horizontal rule.
> 
> Implement truly inline blocks based on the links (I will try to take a look at details later).
> Then we also need some events/hooks as @amantoux suggested for insertion/deletion rules.
> 
> For insertion/deletion rules it would be nice to ask the developer: "Hey, this text is going to be removed/added, here is the text below and after. What shall we do?" And then developer either continues or aborts the change. As a coincidence these customizable rules would solve both (1) and (2).

As you might have guessed I'm in favor of the second approach :). And yes, we need to enrich & package API with customizations (including embeds and rules)

amantoux:
A third way to implement this: enrich SpanEmbed with an `exclusive` (or other name) boolean attribute and a new set of rule to account for this attribute.
This remove the necessity of a configuration/customization API (which will eventually be required but adds complexity)

@zaynetro let me know your thoughts

zaynetro:
I am all for uniformity so if we can manage to implement all of it with `SpanEmbed` I am up to it.

> Yes, solution is the same as implementing SpanEmbed (happens I found a way to solve the height issue)

Could you elaborate more?

I will be able to continue looking after a couple of days.

amantoux:
> > Yes, solution is the same as implementing SpanEmbed (happens I found a way to solve the height issue)

> Could you elaborate more?

`SpanEmbed` are wrapped in a `WidgetSpan` that is added to the `TextSpan` sent to the `TextPainter`
Flutter Quill proceeds the same way is seems

zaynetro:
oh I didn't notice that you have committed but I discovered the same fix with StrutStyle just by tinkering with the code.

Ok so now inline images work and I have discovered a rather decent way how to include custom rules. There was already a good mechanism in place. Thanks for that!

Issues I have noticed:

* Arrow up key doesn't work on the line with images
* Without a custom rule if you insert text after one image then second image will wrap on the second line (Maybe it is ok?)

packages/fleather/lib/src/widgets/text_line.dart
```diff
@@ -166,8 +169,10 @@ class _TextLineState extends State<TextLine> {
 
   InlineSpan _segmentToTextSpan(Node segment, FleatherThemeData theme) {
     if (segment is EmbedNode) {
-      return WidgetSpan(
-          child: EmbedProxy(child: widget.embedBuilder(context, segment)));
+      // TODO: Why do we need EmbedProxy? Things work fine without it.
+      // return WidgetSpan(
+      //     child: EmbedProxy(child: widget.embedBuilder(context, segment)));
+      return WidgetSpan(child: widget.embedBuilder(context, segment));
```
zaynetro, line 175: I couldn't notice the difference here with or without `EmbedProxy`.

zaynetro:
I don't really know why arrow up doesn't work on that line on Linux. Any help is appreciated.

amantoux:
> I don't really know why arrow up doesn't work on that line on Linux. Any help is appreciated.

Unfortunately, I'm not on Linux
@cgestes you're on Linux right? Do you think you can assist?

zaynetro:
I can see the same issue on macOS. If you move cursor to the line with two images then left, right, down arrow keys work but arrow up key doesn't move the cursor at all.

amantoux:
Ok let me check

cgestes:
I believe what you tried to achieve originally and inline images are not exactly the same.

We probably want to be able to support both usecases.

I find the (original) idea of being able to group blocks together very modern and smart.

We probably can achieve the same with inlineSpan and some custom rules, but rules are pretty hard to get right and especially hard to customize. I would assume your original behavior could be behind a block flag to allow / disallow the behavior.

Having images inline is another type of editors, more inline with Word & Google doc.

Have a look at: getTextPositionAbove, getTextPositionBelow for your issue.

Note that cursor handling when we are displaying something else than text needs some polish. If you have hard cases I can give it a look. I am starting to know the code pretty well.

packages/fleather/lib/src/widgets/text_line.dart
cgestes: EmbedProxy is useful for selection. is it the reason you cannot move without it?

cgestes: Play with selecting the content and check the cursor size also around your image with/without.

amantoux:
Re the cursor height, in Flutter, the cursor height is on based on TextStyle of `TextPainter` (using `Paragraph::getBoxesForRange` with `BoxHeightStyle.struct` (see https://github.com/flutter/flutter/blob/76bb8ead5ab1db21cf789c82658ed73071f0273f/packages/flutter/lib/src/painting/text_painter.dart#L976).
We can ask Flutter for the possibility to override the `BoxHeightStyle`

Then there is a strange handling of the resulting caret height for Apple devices in the `CursorPainter::paint`
```dart
switch (defaultTargetPlatform) {
  case TargetPlatform.iOS:
  case TargetPlatform.macOS:
    final double heightDiff = caretHeight - caretRect.height;
    // Center the caret vertically along the text.
    caretRect = Rect.fromLTWH(
      caretRect.left,
      caretRect.top + heightDiff / 2,
      caretRect.width,
      caretRect.height,
    );
    break;
  case TargetPlatform.android:
  case TargetPlatform.fuchsia:
  case TargetPlatform.linux:
  case TargetPlatform.windows:
    // Override the height to take the full height of the glyph at the TextPosition
    // when not on iOS. iOS has special handling that creates a taller caret.
    // TODO(garyq): See the TODO for _computeCaretPrototype().
    caretRect = Rect.fromLTWH(
      caretRect.left,
      caretRect.top - _kCaretHeightOffset,
      caretRect.width,
      caretHeight,
    );
    break;
}
```

I guess we need a separate PR for that

cgestes:
My own take: I believe both use cases make sense and the original behavior is probably welcome by default and very sane (hard to break).

zaynetro:
Thanks for the pointers!

Re: arrow up key not working

The problem is this line: https://github.com/fleather-editor/fleather/blob/6c82fd67be7b990b938f57989b9b836c56772387/packages/fleather/lib/src/rendering/editable_text_line.dart#L293
When caret is on the line with inline images then `preferredLineHeight=21.0` while `height=304.0`.  So we jump up to the same position :)
It seems we need to find a height of the current line, not preferred height nor paragraph's height. The line height might be different depending on whether image is present in the current line or wrapped to the next one.

Unfortunately, I am out of ideas how to fix this ...

packages/fleather/lib/src/widgets/text_line.dart
zaynetro: To me it works the same with and without `EmbedProxy` :shrug: 
I have reverted the change anyway but something you might play with.

zaynetro:
What do you think of the plan:

1. Limit this PR to the ability to extend rules and potentially fixing arrow up key if a good solution is found.
2. I will create a separate PR with my original idea implemented.

I would like to complete (1) so there is a least a solution I can use in my project. As for (2) I am up for discussing it more and would like to hear also @amantoux 's opinion. Frankly, I would love to have (2) implemented.

cgestes:
nice investigation... shouldn't the preferred line height be 304?

We should probably have a look at how it is done in flutter itself. Maybe checking the placeholders dimensions? using one of the method of https://api.flutter.dev/flutter/painting/TextPainter-class.html ?

I'll take some time when I can, to see how it is done in flutter itself.

zaynetro:
>  shouldn't the preferred line height be 304?

This height is with just two images that fit on a single line. If you add more images to the same line then they will soft-wrap. In that case size.height is 624.

amantoux:
> Re the cursor height, in Flutter, the cursor height is on based on `TextStyle` of `TextPainter` (using `Paragraph::getBoxesForRange` with `BoxHeightStyle.struct` (see https://github.com/flutter/flutter/blob/76bb8ead5ab1db21cf789c82658ed73071f0273f/packages/flutter/lib/src/painting/text_painter.dart#L976).
We can ask Flutter for the possibility to override the `BoxHeightStyle` - setting it to BoxHeightStyle.max seems to provide the desired behavior

I believe it all related to the above. Haven't found a way to workaround this...

> What do you think of the plan:
> 1. Limit this PR to the ability to extend rules and potentially fixing arrow up key if a good solution is found.
> 2. I will create a separate PR with my original idea implemented.

Given the above, I think we have to do both indeed

zaynetro:
OK. I don't think I will be able to fix the arrow up in this PR. Ready for review.

zaynetro:
What do you think of these changes?

amantoux:
Hi @zaynetro, I'll take a look this week. A little busy lately, sorry

packages/fleather/example/lib/main.dart
```diff
@@ -113,3 +141,41 @@ class _HomePageState extends State<HomePage> {
...
+    final cursorAfterBlockEmbed =
+        previous != null && _isInlineImage(previous.data);
+
+    if (cursorBeforeBlockEmbed || cursorAfterBlockEmbed) {
+      final delta = Delta()..retain(index);
+      if (cursorAfterBlockEmbed && !data.startsWith('\n')) {
+        delta.insert('\n');
+      }
+      delta.insert(data);
+      if (cursorBeforeBlockEmbed && !data.endsWith('\n')) {
```
amantoux, line 163: Shouldn't this apply if cursorAfterBlockEmbed isn't not verifed?

packages/fleather/example/lib/main.dart
```diff
@@ -113,3 +141,41 @@ class _HomePageState extends State<HomePage> {
...
+
+class ForceNewlineForInsertsAroundInlineImageRule extends InsertRule {
+  @override
+  Delta? apply(Delta document, int index, Object data) {
+    if (data is! String) return null;
+
+    final iter = DeltaIterator(document);
+    final previous = iter.skip(index);
+    final target = iter.next();
+    final cursorBeforeBlockEmbed = _isInlineImage(target.data);
```
amantoux, line 153: Although only in example, maybe rename to `cursorBeforeInlineEmbed`?

amantoux (changes requested):
Thank you

Some small changes in the example to make it work as expected

packages/fleather/example/lib/main.dart
zaynetro: Renamed

packages/fleather/example/lib/main.dart
zaynetro: I am not sure what you mean by that.  

We need both conditions in case user is typing text in-between two inline images.

amantoux: With the current implementation, if both cursorBefore & cursorAfter are verified (typing text in between embeds), the rule inserts new line twice

zaynetro: Yes. If user types text in-between two images we want to insert newline after an image, then insert text and finally insert a new line before second image.

The only problem in this case is that ideally, we would move cursor by one position back.

amantoux: Right, my bad

zaynetro:
Thank for you your time!

Are you planning on making a new release? It would be nice to have this included and [embed height fix](https://github.com/fleather-editor/fleather/pull/68).

amantoux:
I was waiting for your other PR to be done, but if you think it'll take some time, I can release now

zaynetro:
Yeah, the other one will take some time. I want to play more with embeds.

PR #69, opened 2022-12-12
Fixes for the new ParchmentHtmlEncoder

dsyrstad:
Fixes to _ParchmentHtmlEncoder:
1. `ParchmentHtmlCodec` wasn't exported from the package. :smiley: 
1. Text with empty line breaks did not render as blank lines.
2. Text with HTML characters such as < > and & were not escaped.
3. Code blocks (&lt;pre>) were not rendering with line breaks.
4. Block quotes (&lt;blockquote>) were not rendering in the fashion that you see in Fleather, or that you'd typically see in HTML email. 
5. Checklist changes - Added some space between the checkbox and the label, they rendered kind of jammed together. Also, render the checkbox as disabled so the user cannot check/uncheck it.
6. There was an issue with multi-level lists with a plain paragraph following them - the paragraph would get rendering in the list. 
7. Render images so that they will fit within the parent's width, if any. If the parent doesn't have any width, this does nothing. This is useful when generation email HTML, for example, so that the image doesn't bleed out of view.
8. Fix checklist followed by a paragraph.

cc: @amantoux

### DIFF
--- a/packages/parchment/lib/src/convert/html.dart
+++ b/packages/parchment/lib/src/convert/html.dart
@@ -641,8 +641,7 @@ class _HtmlLineTag extends _HtmlTag {
           openTag += '<blockquote$attribute$css>';
         }
         if (attr.value == ParchmentAttribute.code.value) {
-          // We are in a <pre> block at this point.
-          openTag += '<code>';
+          // We are in a <pre><code> block at this point, so no need for an additional <code>.
         }
         if (attr.value == ParchmentAttribute.ol.value) {
           openTag += '<li$attribute$css>';
@@ -678,8 +677,8 @@ class _HtmlLineTag extends _HtmlTag {
           closeTag += '</blockquote>';
         }
         if (attr.value == ParchmentAttribute.code.value) {
-          // We are in a <pre> block. We need to add a newline to display as a line break.
-          closeTag += '</code>\n';
+          // We are in a <pre><code> block. We need to add a newline to display as a line break.
+          closeTag += '\n';
         }
         if (attr.value == ParchmentAttribute.ol.value) {
           closeTag += '</li>';
@@ -745,7 +744,7 @@ class _HtmlBlockTag extends _HtmlTag {
     for (final attr in style.values) {
       if (attr.key == ParchmentAttribute.block.key) {
         if (attr.value == ParchmentAttribute.code.value) {
-          openTag += '<pre>';
+          openTag += '<pre><code>';
         }
         if (attr.value == ParchmentAttribute.ol.value) {
           openTag += '<ol>';
@@ -767,7 +766,7 @@ class _HtmlBlockTag extends _HtmlTag {
     for (final attr in style.values) {
       if (attr.key == ParchmentAttribute.block.key) {
         if (attr.value == ParchmentAttribute.code.value) {
-          openTag += '</pre>';
+          openTag += '</code></pre>';
         }
         if (attr.value == ParchmentAttribute.ol.value) {
           openTag += '</ol>';
--- a/packages/parchment/lib/src/convert/html.dart
+++ b/packages/parchment/lib/src/convert/html.dart
@@ -31,7 +31,7 @@ const _indentWidthInPx = 32;
 /// - default -> <p>
 /// - heading X -> <hX>
 /// - bq -> <blockquote>
-/// - code -> <pre><code>
+/// - code -> <pre>
 /// - ol -> <ol><li>
 /// - ul -> <ul><li>
 /// - cl -> <div class="checklist">
@@ -89,6 +89,10 @@ class _EncoderState {
 // These behave almost as line tags except there can be nested blocks
 class _ParchmentHtmlEncoder extends Converter<Delta, String> {
   const _ParchmentHtmlEncoder();
+
+  static const _htmlElementEscape = HtmlEscape(HtmlEscapeMode.element);
+  static final _brPrEolRegex = RegExp(r'<br></p>$');
+  static final _brEolRegex = RegExp(r'<br>$');
 
   // Style has only positioning attributes
   static bool isPlain(ParchmentStyle style) {
@@ -160,9 +164,7 @@ class _ParchmentHtmlEncoder extends Converter<Delta, String> {
   @override
   String convert(Delta input) {
     final state = _EncoderState();
-    final inputOperations = input.toList();
-    for (var i = 0; i < inputOperations.length; i++) {
-      final op = inputOperations[i];
+    for (final op in input.toList()) {
       final buffer = state.buffer;
       final openInlineTags = state.openInlineTags;
 
@@ -219,11 +221,16 @@ class _ParchmentHtmlEncoder extends Converter<Delta, String> {
       _writeBlockTag(buffer, blockTag..closingPosition = buffer.length);
     }
 
-    // remove default paragraph block if single ligne of text
+    // Remove default paragraph block if single line of text
     String result = buffer.toString();
     if (state.isSingleLine && result.startsWith('<p>')) {
-      return result.substring('<p>'.length, result.length - '</p>'.length);
+      result = result.substring('<p>'.length, result.length - '</p>'.length);
     }
+
+    // Remove the final <br> if there is one.
+    result = result
+        .replaceFirst(_brPrEolRegex, '</p>')
+        .replaceFirst(_brEolRegex, '');
     return result;
   }
 
@@ -321,7 +328,8 @@ class _ParchmentHtmlEncoder extends Converter<Delta, String> {
       position = buffer.length;
     }
 
-    // This assert cannot exist otherwise nested lists won't work.
+    // TODO This assert currently fails the 'Multi-level lists with trailing paragraph' test.
+    // TODO Taking the assert out causes the test to at least generate HTML, albeit incorrect.
     // assert(openBlockTags.length <= 1,
     //     'At most one paragraph should be pushed in stack');
     state.nextLineStartPosition = position;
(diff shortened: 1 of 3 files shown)


amantoux:
Thank you for the PR
I'll try to go through it today

packages/parchment/lib/src/convert/html.dart
```diff
@@ -31,7 +31,7 @@ const _indentWidthInPx = 32;
 /// - default -> <p>
 /// - heading X -> <hX>
 /// - bq -> <blockquote>
-/// - code -> <pre><code>
+/// - code -> <pre>
```
amantoux, line 34: `<pre><code>` was decided according to this
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/code#notes

To represent multiple lines of code, wrap the <code> element within a `<pre>` element. The <code> element by itself only represents a single phrase of code or line of code.

packages/parchment/lib/src/convert/html.dart
```diff
@@ -321,8 +328,10 @@ class _ParchmentHtmlEncoder extends Converter<Delta, String> {
       position = buffer.length;
     }
 
-    assert(openBlockTags.length <= 1,
-        'At most one paragraph should be pushed in stack');
+    // TODO This assert currently fails the 'Multi-level lists with trailing paragraph' test.
```
amantoux, line 331: Shouldn't we try to fix this?

amantoux (changes requested):
I have doubt about how code translate to html with only a `<pre>`
We need to tackle the TODOs

packages/parchment/lib/src/convert/html.dart
dsyrstad: Right - so I'm just using `<pre>` without nested `<code>` elements. I also kept `\n` characters in the `pre` which weren't there before. The result without the `code` element renders the same. For example, go to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/pre and within the "Try It" box, paste this:

```html
<pre >
  @override
  String convert(Delta input) {
    final state = _EncoderState();
    final inputOperations = input.toList();
    for (var i = 0; i &lt; inputOperations.length; i++) {
      final op = inputOperations[i];
    for (final op in input.toList()) {
      final buffer = state.buffer;
      final openInlineTags = state.openInlineTags;
</pre>
```
Notice how it is rendered. Now paste this:
```html
<pre >
<code>  @override</code>
<code>  String convert(Delta input) {</code>
<code>    final state = _EncoderState();</code>
<code>    final inputOperations = input.toList();</code>
<code>    for (var i = 0; i &lt; inputOperations.length; i++) {</code>
<code>      final op = inputOperations[i];</code>
<code>    for (final op in input.toList()) {</code>
<code>      final buffer = state.buffer;</code>
<code>      final openInlineTags = state.openInlineTags;</code>
</pre>
```
You'll see they're exactly the same. The former is just less verbose.

packages/parchment/lib/src/convert/html.dart
dsyrstad: We should and I attempted to, but I don't quite understand how the code is handling it. I think we should probably close a block if the indent level changes, or the block type. I could take another look at this today. Any hints you may have would be appreciated.

packages/parchment/lib/src/convert/html.dart
```diff
@@ -608,7 +641,7 @@ class _HtmlLineTag extends _HtmlTag {
           openTag += '<blockquote$attribute$css>';
         }
         if (attr.value == ParchmentAttribute.code.value) {
-          openTag += '<code>';
+          // We are in a <pre> block at this point. No <code> tag is needed.
```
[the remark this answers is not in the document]
dsyrstad: See above.

packages/parchment/lib/src/convert/html.dart
amantoux: I did some reading 

The `<pre>` [HTML](https://developer.mozilla.org/en-US/docs/Web/HTML) element represents preformatted text which is to be presented exactly as written in the HTML file. The text is typically rendered using a non-proportional, or [monospaced](https://en.wikipedia.org/wiki/Monospaced_font), font. Whitespace inside this element is displayed as written.

The `<code>` [HTML](https://developer.mozilla.org/en-US/docs/Web/HTML) element displays its contents styled in a fashion intended to indicate that the text is a short fragment of computer code. By default, the content text is displayed using the [user agent's](https://developer.mozilla.org/en-US/docs/Glossary/User_agent) default monospace font.

So it we want to display code we should use `<code>` and since we want to preserve format of text (new lines, spaces) we should embed it in `<pre>`

```html
<pre>
<code>
@override
  String convert(Delta input) {
    final state = _EncoderState();
    final inputOperations = input.toList();
    for (var i = 0; i &lt; inputOperations.length; i++) {
      final op = inputOperations[i];
    for (final op in input.toList()) {
      final buffer = state.buffer;
      final openInlineTags = state.openInlineTags;
<code>
</pre>

Does it make sense?

Comment (dsyrstad) [reply packages/parchment/lib/src/convert/html.dart:34]: Ok, it is a semantic difference, but I'll take care of it.

File: packages/parchment/lib/src/convert/html.dart (hunk 2 above)
Comment (amantoux) [reply packages/parchment/lib/src/convert/html.dart:331]: prior to handling the plain block, we handle any previously opened block. In this instance the stack is not being emptied - only the last inserted element (list with indent = 1) was popped, the list element with no indent remains.
Later, the paragraph block is added to the stack (thus more than one)

File: packages/parchment/lib/src/convert/html.dart (hunk 1 above)
Comment (dsyrstad) [reply packages/parchment/lib/src/convert/html.dart]: Done

**Review by amantoux — CHANGES_REQUESTED** [review]

File: packages/parchment/test/convert/html_test.dart [hunk 4]
```diff
@@ -256,11 +256,10 @@ void main() {
 
         expect(
           codec.encode(doc.toDelta()),
-          '<pre>'
-          'void main() {\n'
-          '\n'
-          '  print("Hello World!");\n'
-          '}\n'
+          '<pre><code>void main() {</code>\n'
```
amantoux, line 259: There might be some misunderstanding
Your previous implementation was correct only it needed the whole block to be `<pre><code>` and only the block
```html
<pre><code>
void main() {
 
  println(...)
}
</code></pre>
```

and **not**

```html
<pre>
<code>void main() {</code>
<code>\n</code>
<code>  println(...)</code>
<code>}</code>
</pre>
```

dsyrstad: Changed.

packages/parchment/lib/src/convert/html.dart
dsyrstad: I'm working on this...

dsyrstad: This is fixed. It took quite a bit of time to figure out the proper way to close blocks that would work for everything.

packages/parchment/lib/src/convert/html.dart
```diff
@@ -205,26 +211,43 @@ class _ParchmentHtmlEncoder extends Converter<Delta, String> {
...
+
+    // Remove the final <br> if there is one.
+    result = result
+        .replaceFirst(_brPrEolRegex, '</p>')
+        .replaceFirst(_brEolRegex, '');
+    return result;
+  }
+
+  /// Closes all open blocks and returns the ending position.
+  int _closeOpenBlocks(_EncoderState state) {
```
amantoux, line 230: This only works for checklist followed by paragraph but doesn't work for other blocks followed by paragraph.
I've come up with this
```dart
int _closeOpenBlocks(_EncoderState state,
      {bool beforePlainParagraphHandling = false}) {
    final openBlockTags = state.openBlockTags;
    final buffer = state.buffer;
    final numToClose = openBlockTags.length;
    int position = 0;
    for (var i = 0; i < numToClose; i++) {
      final blockTag = openBlockTags[i];
      if (position > 0) {
        blockTag.closingPosition = position;
      } else {
        position = blockTag.closingPosition;
      }
      if (!isPlain(blockTag.style)) {
        position += blockTag.inducedPadding;
      }
      if (i == numToClose - 1 && !beforePlainParagraphHandling) {
        blockTag.closingPosition = buffer.length;
      }
      _writeBlockTag(buffer, blockTag);
    }

    state.openBlockTags.clear();
    return numToClose == 1 ? position : buffer.length;
  }
```

for the last opened tags, if a paragraph follows, we will be using the blockTag's closing position.

Also there were problems with nested block, the position variable is updated to handle this

packages/parchment/lib/src/convert/html.dart
```diff
@@ -282,13 +305,9 @@ class _ParchmentHtmlEncoder extends Converter<Delta, String> {
...
     final buffer = state.buffer;
+
     if (openBlockTags.isNotEmpty) {
-      final currentBlock = openBlockTags.removeAt(0);
-      position = currentBlock.closingPosition;
-      if (!isPlain(currentBlock.style)) {
-        _writeBlockTag(buffer, currentBlock);
-        position += currentBlock.inducedPadding;
-      }
+      position = _closeOpenBlocks(state);
```
amantoux, line 310: call with `beforeParagraphHandling: true `

packages/parchment/test/convert/html_test.dart
```diff
@@ -74,6 +74,77 @@ void main() {
...
+          },
+          {'insert': '\n \n\n\n'},
+          {
+            'insert': 'Line 6',
+            'attributes': {'b': true}
+          },
+          {'insert': '\n'}
+        ]);
+        expect(codec.encode(doc.toDelta()),
+            '<p>Line 1</p><p><br></p><p><br></p><p><br></p><p>Line 2</p><p>Line3</p><p>not blank1</p><p>not blank2</p><p>not blank3</p><p>Line 4</p><p><strong>Line 5</strong></p><p> <br></p><p><br></p><p><br></p><p><strong>Line 6</strong></p>');
```
amantoux, line 113: this is quite a long line, can we split for more readability?

packages/parchment/test/convert/html_test.dart
```diff
@@ -74,6 +74,77 @@ void main() {
...
+          },
+          {'insert': '\n'},
+          {
+            'insert': 'Bold',
+            'attributes': {'b': true}
+          },
+          {'insert': '\n'},
+        ]);
+        expect(codec.encode(doc.toDelta()),
+            '<p><strong>Bold</strong></p><p><em>Italic</em></p><p><strong>Bold</strong></p><p><em>Italic</em></p><p><strong>Bold</strong></p>');
```
amantoux, line 146: same readability issue

packages/parchment/lib/src/convert/html.dart
dsyrstad: OK, all the tests were passing with my version. Perhaps you need a test to demonstrate your failing case? Nevertheless, I put your version in and all of the still pass.

amantoux: Yes this failed
```json
{
  "input":[
    {"insert": "some code"},
    {
      "insert": "\n",
      "attributes": {"block": "code"}
    },
    {"insert": "Hello world", "attributes":{"b": true}},
    {"insert": "\n"}
  ]
}
```
```dart
'<pre><code>some code\n</code></pre><p><strong>Hello world</strong></p>'
```

dsyrstad: Great! I've added a test.

amantoux (approved):
LGTM
Thanks!

PR #73, opened 2022-12-16
enable getBoxesForSelection

cgestes:
trivial.

I use it to calculate proper position for a hover on mentions spanning multiple lines.


amantoux (approved):
LGTM

Thanks

PR #74, opened 2022-12-16
more graceful error handling

cgestes:
- document: throw instead of assert when document is invalid, invalid document is just a valid error, leave a chance to the user of the API to fix it.
- display an error instead of throwing on invalid textspan / embed.
Another fix may be avoid throwing on in the default embed builder.

what do you think?


amantoux:
> * document: throw instead of assert when document is invalid, invalid document is just a valid error, leave a chance to the user of the API to fix it.

Yes sounds better

> * display an error instead of throwing on invalid textspan / embed.
>   Another fix may be avoid throwing on in the default embed builder.

Rather than prompting the entire error in a `TextSpan`, what about prompting a generic message and log in console more technical message with error description?

> what do you think?

cgestes:
Ok, will do that.

PR #99, opened 2023-02-14
Fix wrong toolbar positioning

### LINKED ISSUE #96: Selection toolbar wrong positioning
## Steps to Reproduce

1. Run the example app
2. Select the heading text ("Fleather")
3. Instead of displaying it right above or under the selected text, it appears way below it.

Amir-P:
Fixes #96 using Flutter's `TextSelectionToolbarAnchors` to compute contextMenuAnchors.


amantoux (commented):
neater! Do you think we could detect such issues in some tests?

Amir-P:
> @Amir-P Neater! Do you think we could detect such issues in some tests?

I'm not expert in widget testing (yet!) and not sure how the tests might look like for this. Do you have anything in mind? @amantoux

Amir-P:
I think the only way to validate toolbar positioning is using golden tests which is a bit tricky to get it work correctly since there are differences in text rendering on different platforms and generated files on my machine (macOS) is not the same as the ones generated in our CI pipeline (Linux). I think we can skip the tests for this for now. WDYT? @amantoux

amantoux:
> I think the only way to validate toolbar positioning is using golden tests which is a bit tricky to get it work correctly since there are differences in text rendering on different platforms and generated files on my machine (macOS) is not the same as the ones generated in our CI pipeline (Linux). I think we can skip the tests for this for now. WDYT? @amantoux

Ok then, let's merge this

PR #113, opened 2023-04-03
Use document in codecs

Amir-P:
## Changes
* Changes codecs to use `ParchmentDocument` instead of `Delta`
* Fixes issues with markdown decoder.

## Why?
By using `ParchmentDocument` we ensure standard input for encoders and output for decoders.


Amir-P:
Alan can you please review? Also any idea why codecov fails? @amantoux

amantoux (approved):
LGTM

PR #125, opened 2023-05-31
Cleaning up legacy files and adding LICENSE to repository root

amantoux:


amantoux:
@Amir-P PTAL

Amir-P (commented):
Given that I have little knowledge of copyright and software licensing, I have two questions:
1. Do you consider Fleather a redistribution of Zefyr?
2. If so, shouldn't we mention previous copyright, as it's clearly stated by the previous license?

amantoux:
> Given that I have little knowledge of copyright and software licensing, I have two questions:
> 
> 1. Do you consider Fleather a redistribution of Zefyr?
> 2. If so, shouldn't we mention previous copyright, as it's clearly stated by the previous license?

Don't know how important this is, I'll add it in case

PR #137, opened 2023-06-15
Migrate to Dart 3

amantoux:
Migrate code to Dart 3

### DIFF
--- a/packages/fleather/example/pubspec.yaml
+++ b/packages/fleather/example/pubspec.yaml
@@ -6,7 +6,7 @@ version: 1.0.0+1
 publish_to: none
 
 environment:
-  sdk: ">=2.15.0 <3.0.0"
+  sdk: ">=3.0.0 <4.0.0"
 
 dependencies:
   flutter:
--- a/packages/fleather/lib/src/rendering/editor.dart
+++ b/packages/fleather/lib/src/rendering/editor.dart
@@ -982,7 +982,7 @@ class RenderEditor extends RenderEditableContainerBox
   }
 }
 
-class FleatherVerticalCaretMovementRun extends Iterator<TextPosition> {
+class FleatherVerticalCaretMovementRun implements Iterator<TextPosition> {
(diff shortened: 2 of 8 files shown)


Amir-P (commented):
I have some questions since I haven't worked with Dart 3 yet:

1. Why did you define some classes with `final` modifier? Is there any reason for restricting their extensibility?
2. Does it force dependees to upgrade to Dart 3?

amantoux:
1. Because I believe we should be responsible of the `Parchment` model not the user of the API. If it weren't the case, a user could define a new Node type that could not be handle by Fleather any differently from it's parent Node.
I'm event considering using a sealed class :)
3. Quoting https://dart.dev/resources/dart-3-migration#unversioned-vs-versioned-changes
> Dart 3 backwards compatibility
>
>Many packages and apps that used null safety with Dart 2.12 or later are likely backwards compatible with Dart 3. This is possible for any package where the lower bound of the SDK constraint is 2.12.0 or higher.
>
>[Dart’s pub tool](https://dart.dev/guides/packages) allows resolution even when the upper bound is limited to versions below 3.0.0. For example, a package with the following constraint will be allowed to resolve with a Dart 3.x SDK, as pub will re-interpret the upper-constraint <3.0.0 as <4.0.0 when the lower constraint is 2.12 or higher:
>
>environment:
  sdk: '>=2.14.0 <3.0.0'           # This is interpreted as '>=2.14.0 <4.0.0'
>
>This allows developers to use Dart 3 sound null safety with packages that already support 2.12 null safety without needing a second migration, unless the code is affected by any other Dart 3 changes.

Amir-P:
> 1. Because I believe we should be responsible of the `Parchment` model not the user of the API. If it weren't the case, a user could define a new Node type that could not be handle by Fleather any differently from it's parent Node.

I'm more concerned with Parchment and Fleather extensibility. I can't think of any case now. We can change it later when something comes up.

> @Amir-P Can you give it try?

Seems to be working.

Let's merge it.

PR #142, opened 2023-06-19
Revert "Upload to codecov lies in its own job"

amantoux:
Reverts fleather-editor/fleather#140


Amir-P:
What was the reason for reverting it @amantoux ?

amantoux:
> What was the reason for reverting it? @amantoux

@Amir-P , I was about to comment this.
Actually, the decoupling isn't sending properly the cov files to Codecov, it need more investigation

PR #77, opened 2022-12-18
Ensure that text boxes are only the required size

cgestes:
Otherwise hitTest are wrong and onEnter/onExit/recognizer/mouseCursor of TextSpans recognise the trailing empty space as a hit while it actually is not.

## Before:
https://user-images.githubusercontent.com/49141/208324135-26cd552d-403f-41e3-a554-469c7b183e6a.mov

image

## After:
https://user-images.githubusercontent.com/49141/208324132-a03ce4f7-f1d0-4af4-92d7-be37f2bca324.mov

image

PR #144, opened 2023-06-21
Fix wrong bounding box for text line in block

### LINKED ISSUE #143: Different bounding box for text line and box
Text bounding box for text box is computed differently from text lines. This also causes issues with text alignment and direction for text lines in blocks.

Screenshot 2023-06-21 at 09 24 29

Amir-P:
Fixes #143


amantoux:
@Amir-P I think this break the behavior designed in PR https://github.com/fleather-editor/fleather/pull/77

Amir-P:
> @Amir-P I think this break the behavior designed in PR https://github.com/fleather-editor/fleather/pull/77

Can't we address that issue when hit testing? Also why it's not a problem with outside of block text lines? @amantoux

amantoux:
> > @Amir-P I think this break the behavior designed in PR #77
> 
> Can't we address that issue when hit testing? Also why it's not a problem with outside of block text lines? @amantoux

Hey @Amir-P, were you able to dig a little further on this?

Amir-P:
> Hey @Amir-P, were you able to dig a little further on this?

Not yet. I'll get to it tomorrow. Will let you know. @amantoux

Amir-P:
There is definitely a problem with hit testing but the fix at #77 is only solving problem for blocks and causes another issue (#143). We need to revert that and then find a way for fixing underlying problem. WDYT? @amantoux 

A normal text line with link on master branch showing the hit test problem:

Screenshot 2023-07-04 at 09 48 21

Amir-P:
BTW, instead of reverting #77 I prefer to merge the code in this PR since it's not forcing it to take max width.

amantoux:
@Amir-P Either solution works but if 2. is easily acheivable, I think we should go for this one

Amir-P:
> @Amir-P Either solution works but if 2. is easily acheivable, I think we should go for this one

Yeah, after experimenting with `RichText` and `TextPainter`, changes for option 1 is a lot. I think option 2 is safer. I'll push another commit on this branch to make lines bounding box like blocks and will update the issue description too. Then will fix the alignment issue.

Amir-P:
Can you take a look please? @amantoux

Amir-P:
Finally, I've decided to keep text lines constraints unchanged and instead check if hit test position is in text boxes. I think it's ready for review. @amantoux

amantoux (approved):
LGTM! Thanks @Amir-P

PR #179, opened 2023-12-01
Show context menu on desktop

Amir-P:


packages/fleather/lib/src/widgets/text_selection.dart
```diff
@@ -660,176 +879,777 @@ class EditorTextSelectionGestureDetectorBuilder {
...
   /// By default, it selects word edge if selection is enabled.
   ///
   /// See also:
   ///
-  ///  * [EditorTextSelectionGestureDetector.onSingleTapUp], which triggers
+  ///  * [TextSelectionGestureDetector.onSingleTapUp], which triggers
   ///    this callback.
   @protected
-  void onSingleTapUp(TapUpDetails details) {
+  void onSingleTapUp(TapDragUpDetails details) {
```
amantoux, line 905: What is the rational behing the change of details type?

packages/fleather/lib/src/widgets/text_selection.dart
```diff
@@ -660,176 +879,777 @@ class EditorTextSelectionGestureDetectorBuilder {
...
+            final TextSelection? fromSelection = renderEditor.hasFocus
+                ? null
+                : const TextSelection.collapsed(offset: 0);
+            _expandSelection(
+              details.globalPosition,
+              SelectionChangedCause.tap,
+              fromSelection,
+            );
+            return;
+          }
```
amantoux, line 942: Guessing this is as implemented in Flutter right?

packages/fleather/lib/src/widgets/text_selection.dart
```diff
@@ -660,176 +879,777 @@ class EditorTextSelectionGestureDetectorBuilder {
...
   ///
   /// See also:
   ///
-  ///  * [EditorTextSelectionGestureDetector.onDragSelectionStart], which triggers
+  ///  * [TextSelectionGestureDetector.onDragSelectionStart], which triggers
   ///    this callback.
   @protected
-  void onDragSelectionStart(DragStartDetails details) {
-    renderEditor!.handleDragStart(details);
+  void onDragSelectionStart(TapDragStartDetails details) {
```
amantoux, line 1296: Not sure I want to go through all this in details :)

amantoux (commented):
Great to see this feature!
Some comments & questions, but overall ok

Amir-P:
> @Amir-P Did you run some manual tests on iOS and Android? The changes impact it quite a lot

I only got to test it on mac and iOS emulator. Haven't got to test in on Android and Web. I also don't have access to Linux and Windows machine. This PR has so many potential to introduce unintended behaviors but as you know, it is copied from Flutter's source code with very little modifications. @amantoux

amantoux:
> > @Amir-P Did you run some manual tests on iOS and Android? The changes impact it quite a lot
> 
> 
> 
> I only got to test it on mac and iOS emulator. Haven't tested it on Android and Web yet. I also don't have access to Linux and Windows machine. This PR has so many potential to introduce unintended behaviors but as you know, it is copied from Flutter's source code with very little modifications. @amantoux 

Ok never mind 
We see that in another PR if need be

packages/fleather/lib/src/widgets/text_selection.dart
Amir-P: Before, we were using our own gesture detector but now with `TextSelectionGestureDetector` we have to provide  `GestureTapDragUpCallback` which is a `void Function(TapDragUpDetails details)`. So it should match its signature.

packages/fleather/lib/src/widgets/text_selection.dart
Amir-P: https://github.com/fleather-editor/fleather/pull/179#discussion_r1417839182

packages/fleather/lib/src/widgets/text_selection.dart
Amir-P: Yes. You check here: https://github.com/flutter/flutter/blob/db7ef5bf9f59442b0e200a90587e8fa5e0c6336a/packages/flutter/lib/src/widgets/text_selection.dart#L2317

PR #181, opened 2023-12-03
Auto formats

amantoux:
Following discussion #170, provides implemention of formatting suggestion as the user types in the editor.

## Description
As user types in text, the editor can automatically suggest formatting based on a set of automatic format detection rules.
Such rule can, for example, detect links or detect Markdown shortcut.
- [x] Once activated, formatting suggestions can be canceled or undone only before next local change following this activation.
- [x] In current implement, a deletion of a character at the end of the formatted area will undo and deactivate the formatting suggestion
- [x] Automatic formatting actions are added to the editor's history, the user can therefore undo any auto format results via the history.

### Link detection
Inserting a space at the end of a URI formats the link accordingly and adds hyperlink to URI. 
The `space` is add to the document.
Deleting the `space` cancels the formatting without removing it from the document
Cursor follows usual behavior

### Markdown shortcuts
Inserting a `space` at the end of a shortcut formats the block according.
The space is NOT added to the document
Deleting a character from the start of the resulting bloc cancels the formatting but restores shortcut with added `space`


amantoux:
>     1. It is impossible to write a link with no link attribute. If you press backspace after link auto-formatted, and then press space, it get's auto-formatted again. This is different in Google Docs. Backspace doesn't remove the space after link or any other auto-formatted piece of text but just reverses the formatting.
Agreed, don't how I missed that ^^

>     2. Pressing enter after text direction still removes the test direction attribute from line
I don't think this has to do with the auto format but rather with the insertion after a rtl line

Amir-P:
> Agreed, don't how I missed that ^^. Changed the behavior, can you give it another try?

Is there any case where we want to  reverse the formatting and also remove the last character? I think it's a general thing that backspace immediately after an auto format should only undo the formatting and not remove the character. WDYT? @amantoux 

> I don't think this has to do with the auto format but rather with the insertion after a rtl line

I will investigate this tomorrow. As far as I know, the line style should be preserved via heuristics.

amantoux:
> Is there any case where we want to  reverse the formatting and also remove the last character? I think it's a general thing that backspace immediately after an auto format should only undo the formatting and not remove the character. WDYT? @amantoux

I'm not sure I understand. In the last commit, the link is formatted after the space, if you immediately hit backspace it undoes the formatting without deleting the space character. 

> I will investigate this tomorrow. As far as I know, the line style should be preserved via heuristics.

Yes PreserveBlockStyleInsertRule might need to become a PreserveLineStyleInsertRule

Amir-P:
> Yes PreserveBlockStyleInsertRule might need to become a PreserveLineStyleInsertRule

Looks like the issue comes from this part of. I will work on it after merging this.
https://github.com/fleather-editor/fleather/blob/ad8a0728e6692948920cc587a5ed622c8dbe9ebd/packages/parchment/lib/src/heuristics/insert_rules.dart#L139-L153

Amir-P (approved):
LGTM. Thanks.

PR #187, opened 2023-12-07
Support for automatic mode in color selection buttons

### LINKED ISSUE #172: Handle no color selection for text & background color
By default, colors of background & text change based on the current brightness of the app.
When selecting another color, there is no way to restore the default dynamic colors

Add a button to restore default dynamic colors; `Automatic` for text color & `No color` for background color

amantoux:
Closes #172

### DIFF
--- a/packages/fleather/lib/src/widgets/editor_toolbar.dart
+++ b/packages/fleather/lib/src/widgets/editor_toolbar.dart
@@ -350,8 +350,10 @@ Widget defaultToggleStyleButtonBuilder(
   );
 }
 
-/// Signature of callbacks that return a [Color] picked from a [BuildContext].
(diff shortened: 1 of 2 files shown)


Amir-P (dismissed):
I think "No Color" or "Transparent" or "Remove background color" is a better term than "Automatic" since that's what really happening. WDYT?

amantoux:
> I think "No Color" or "Transparent" or "Remove background color" is a better term than "Automatic" since that's what really happening. WDYT?

I used MS Word wording. I believe the rational behind this is that the color automatically shifts from light mode color to dark mode color. Other possibilities not not have dynamic colors

image

amantoux:
Changed accordingly ✅

Amir-P (approved):
LGTM.

PR #191, opened 2023-12-10
Fix text direction auto format detection and undo

Amir-P:
It fixes two issues:

1. When users enters a character in an empty line with line style auto text direction is not working correctly
2. When auto format applies the same style the undo is an empty delta and applying it causes exception

### DIFF
--- a/packages/fleather/lib/src/widgets/controller.dart
+++ b/packages/fleather/lib/src/widgets/controller.dart
@@ -171,6 +171,8 @@ class FleatherController extends ChangeNotifier {
   // document; `false` otherwise
   bool _captureAutoFormatCancellationOrUndo(
       ParchmentDocument document, int position, int length, Object data) {
+    if (!_autoFormats.hasActiveSuggestion) return true;
+    
     if (_autoFormats.canUndo) {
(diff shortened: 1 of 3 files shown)


Amir-P:
Also there is a small behavioral difference with Google Docs. Google Docs doesn't have an undo for auto text direction and just remove the entered character. Should we change it too? @amantoux

packages/fleather/lib/src/widgets/controller.dart
```diff
@@ -171,7 +171,7 @@ class FleatherController extends ChangeNotifier {
   // document; `false` otherwise
   bool _captureAutoFormatCancellationOrUndo(
       ParchmentDocument document, int position, int length, Object data) {
-    if (!_autoFormats.hasActiveSuggestion) return true;
+    if (!_autoFormats.canUndo) return true;
```
amantoux, line 174: If active suggestions with empty undo, we still need to cancel active suggestion I think

amantoux (changes requested):
PTAL

amantoux (approved):
LTGM

PR #203, opened 2024-01-01
Android - add selection handle on collapsed selection

### LINKED ISSUE #199: [Android] Collapsed selection handle missing
On Android, when selection is collapsed, a single rotated handle should be visible to allow user to move cursor
Targeted behavior when setting cursor by taping at a position
Capture d’écran 2023-12-31 à 12 38 05

amantoux:
Closes #199

### DIFF
--- a/packages/fleather/lib/src/widgets/editor.dart
+++ b/packages/fleather/lib/src/widgets/editor.dart
@@ -1265,11 +1265,6 @@ class RawEditorState extends EditorState
     performSpellCheck(widget.controller.plainTextEditingValue.text);
   }
 
-  bool _shouldShowSelectionHandles() {
-    return widget.showSelectionHandles &&
-        !widget.controller.selection.isCollapsed;
-  }
-
   @override
   void didUpdateWidget(RawEditor oldWidget) {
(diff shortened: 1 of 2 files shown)


packages/fleather/lib/src/widgets/editor.dart
```diff
@@ -1265,11 +1265,6 @@ class RawEditorState extends EditorState
     performSpellCheck(widget.controller.plainTextEditingValue.text);
   }
 
-  bool _shouldShowSelectionHandles() {
```
Amir-P, line 1268: What was the purpose of this function that is removed now?

Amir-P (commented):
Just had a question. PTAL.

packages/fleather/lib/src/widgets/editor.dart
amantoux: It was used to show or hide the handles depending on whether the selection was collapsed or not.
Now the show/hide of the handles is entirely managed by `EditorTextSelectionOverlay` allowing for platform specific behavior

Amir-P (approved):

PR #207, opened 2024-01-05
Fix RenderEditableTextLine's getPositionAbove

### LINKED ISSUE #192: Unexpected behavior of vertical cursor movements from a line with span embeds
## Steps to Reproduce
Vertical cursor movements have incorrect behavior when initiated from a line with large embeds

In the example, put the cursor at the end of line with 2 span embed and hit ↑
--> cursor doesn't move

hit ↑ again
--> cursor moves to beginning of document

Amir-P:
Fixes #192

### DIFF
--- a/packages/fleather/lib/src/rendering/editable_text_line.dart
+++ b/packages/fleather/lib/src/rendering/editable_text_line.dart
@@ -285,18 +285,10 @@ class RenderEditableTextLine extends RenderEditableBox {
   // Computes the box height for the position.
   // To get the correct box, it extends position backward if position
   // is bigger than 0 and extends it forward if it's text position is 0.
-  double _getBoxHeightForPosition(TextPosition position) {
-    final int baseOffset;
-    if (position.offset > 0) {
-      baseOffset = position.offset - 1;
-    } else {
-      baseOffset = position.offset;
-    }
+  double _getLineHeightForPosition(TextPosition position) {
+    final lineBoundary = getLineBoundary(position);
     final boxes = body!.getBoxesForSelection(TextSelection(
-      baseOffset: baseOffset,
-      extentOffset: baseOffset + 1,
-      affinity: position.affinity,
-    ));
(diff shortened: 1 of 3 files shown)


packages/fleather/lib/src/rendering/editable_text_line.dart
```diff
@@ -282,6 +282,24 @@
     return body!.getPositionForOffset(shiftedOffset);
   }
 
+  // Computes the box height for the position.
+  // To get the correct box, it extends position backward if position
+  // is bigger than 0 and extends it forward if it's text position is 0.
+  double _getBoxHeightForPosition(TextPosition position) {
```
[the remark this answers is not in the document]
Amir-P: True. I was thinking about this too while having dinner! Will change it. @amantoux

packages/fleather/lib/src/rendering/editable_text_line.dart
```diff
@@ -285,18 +285,10 @@ class RenderEditableTextLine extends RenderEditableBox {
...
-    }
+  double _getLineHeightForPosition(TextPosition position) {
+    final lineBoundary = getLineBoundary(position);
     final boxes = body!.getBoxesForSelection(TextSelection(
-      baseOffset: baseOffset,
-      extentOffset: baseOffset + 1,
-      affinity: position.affinity,
-    ));
+        baseOffset: lineBoundary.start, extentOffset: lineBoundary.end));
     return boxes.firstOrNull?.toRect().height ?? 0;
```
amantoux, line 292: Don't we rather want to get the max height of all boxes of the line?

amantoux (changes requested):
Hey @Amir-P 
I still have a question/comment. Can you let me know your thoughts?

packages/fleather/lib/src/rendering/editable_text_line.dart
Amir-P: Based on my tests it returns one box but that's better. I will change it now.

amantoux (approved):
LTGM!

PR #210, opened 2024-01-07
Use framework context menu instead on browser context menu on web

amantoux:
Use of browser context menu on web causes some issue that make it unusable. (see https://github.com/flutter/flutter/issues/134024)
This PR proposes to use the context menu provided by the framework instead

### DIFF
--- a/packages/fleather/example/lib/main.dart
+++ b/packages/fleather/example/lib/main.dart
@@ -1,6 +1,7 @@
 import 'dart:convert';
 
 import 'package:fleather/fleather.dart';
+import 'package:flutter/foundation.dart';
 import 'package:flutter/material.dart';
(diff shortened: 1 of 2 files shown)


Amir-P (changes requested):
My concern is this disables the context menu application wide and not editor only. I think it would be better to use introduce a flag to enable/disable this.

amantoux:
> My concern is this disables the context menu application wide and not for editor only. I think it would be better to use introduce a flag to enable/disable this.

Yes agreed. I've delegated the enabling/disabling of the `BrowserContextMenu` to the application host

Amir-P (approved):
Thanks, it's better now. Just that `dispose` can be deleted.
We should add a section to doc to inform users it's possible to use framework context menu on Web.

PR #211, opened 2024-01-07
[web] Update HTML contentEditable with correct size and style

amantoux:
The PR attempts to make proper use of the `TextInputConnection` API in order to better handle cursor movements on web browsers.

### Known issue: 
- "go to line end" sometimes moves cursor to start of next line within a same paragraph
- arrow up keydown from the first line of a paragraph will move to previous paragraph but not necessarily at the right position if paragraphs have different style

### DIFF
--- a/packages/fleather/lib/src/rendering/editable_text_line.dart
+++ b/packages/fleather/lib/src/rendering/editable_text_line.dart
@@ -19,7 +19,6 @@ enum TextLineSlot { leading, body }
 class RenderEditableTextLine extends RenderEditableBox {
   /// Creates new editable paragraph render box.
   RenderEditableTextLine({
-    // RenderEditableMetricsProvider child,
     required LineNode node,
     required EdgeInsetsGeometry padding,
     required TextDirection textDirection,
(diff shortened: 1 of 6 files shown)


amantoux:
@Amir-P It still needs some tests, but can you run some tests and let me know how it goes?

Amir-P (commented):
It works better than before but there are still some issues when used in lines with inline spans. Can we fix it in this PR too?
Also can we introduce a style builder which generates text style based on parchment style and use it both in `text_line.dart` and `editor_input_client_mixin.dart`? It prevents the duplication added in this PR.

amantoux:
> Thanks, Alan! It works better than before but there are still some issues when used in lines with inline spans. Can we fix it in this PR too? Also can we introduce a style builder which generates text style based on parchment style and use it both in `text_line.dart` and `editor_input_client_mixin.dart`? It prevents the duplication added in this PR.

`text_line.dart` is finer than `editor_input_client_mixin.dart` because text style is the same as paragraph style for lists and other blocks

Do you mind explicating the issues with the inline spans?

Amir-P:
@amantoux PTAL at the video:

https://github.com/fleather-editor/fleather/assets/8766492/05b742bb-6d4a-4f68-820e-3ea3004ec369

amantoux:
> @amantoux PTAL at the video:
> fleather.mov

@Amir-P can't figure out how to change this behavior (and I don't think it is that bad?)

Amir-P:
> @Amir-P can't figure out how to change this behavior (and I don't think it is that bad?)

It's not that bad but it is a bit different than the behavior on other platforms. As a user I expect it to land at the same offset as it was before in the previous line. But I can't think of an easy way to fix this. Let's keep it like this for now. @amantoux

Amir-P (approved):
LGTM

PR #213, opened 2024-01-09
Adapt Fleather pubspec to silent pub.dev warnings

amantoux:
Reference published package of `Parchment` rather than locally referencing it


Amir-P (commented):
Wouldn't this break when we make unpublished changes in Parchment and need them in Fleather?

amantoux:
> Wouldn't this break when we make unpublished changes in Parchment and need them in Fleather?

Yes it would mean we need to published Parchment changes prior to publishing, but if we want to keep two separate packages, I can't think of another way to do this properly

Amir-P:
We would have to publish every small changes and have separated PRs for Fleather and Parchment changes. What if we move our development works to another branch like `develop`? Keep the master branch for the latest published version on pub.dev only. @amantoux

amantoux:
> We would have to publish every small changes and have separated PRs for Fleather and Parchment changes. What if we move our development works to another branch like `develop`? Keep the master branch for the latest published version on pub.dev only. @amantoux

Yes it sounds better @Amir-P

amantoux:
I'm trying another solution prior to creating a new branch, let me know you thoughts @Amir-P 

We use `dependencies_overrides` with local Parchment (This override is ignored when publishing if I get it right) and we can remove 'publish_to: none' to remove error from pub.dev

Amir-P (approved):
LGTM! Thanks for doing it.

PR #218, opened 2024-01-10
Add image picker to example

### LINKED ISSUE #212: How to insert image and video or any custom widget
How can we insert an image either through link or an local image in the editor, that to with the image icon in the tool bar, and also is there any chance that if we copied an image from some where and pasted it in the editor, the image should be pasted same as the text.

If so, any example to lead me through?

Amir-P:
Closes #212


amantoux (approved):
LGTM

Issue #227, opened 2024-01-15
Sometimes receive assertion when pressing arrow keys down, and then pressing enter.


nialljawad96:
## Steps to Reproduce

1. Press arrow keys down in fleather field to go lower in the textfield.
2. Press enter a few times.
3. Receive assertion.

This appears to happen particularly when going past (vertically) a text/embed that is already present, with the arrow down key. 

## Environment
- Mac OS
- Flutter version 3.16.4
- Fleather version 1.13.2

## Logs

```

======== Exception caught by services library ======================================================
The following assertion was thrown during method call TextInputClient.updateEditingStateWithDeltas:
'package:parchment/src/document/node.dart': Failed assertion: line 235 pos 12: 'index == 0 || (index > 0 && index < length)': is not true.

When the exception was thrown, this was the stack: 
#2      ContainerNode.insert (package:parchment/src/document/node.dart:235:12)
#3      ParchmentDocument.compose (package:parchment/src/document.dart:246:15)
#4      ParchmentDocument.insert (package:parchment/src/document.dart:123:5)
#5      ParchmentDocument.replace (package:parchment/src/document.dart:162:15)
#6      FleatherController.replaceText (package:fleather/src/widgets/controller.dart:129:24)
#7      RawEditorStateTextInputClientMixin.updateEditingValueWithDeltas (package:fleather/src/widgets/editor_input_client_mixin.dart:159:25)
#8      TextInput._handleTextInputInvocation (package:flutter/src/services/text_input.dart:1870:63)
#9      TextInput._loudlyHandleTextInputInvocation (package:flutter/src/services/text_input.dart:1759:20)
#10     MethodChannel._handleAsMethodCall (package:flutter/src/services/platform_channel.dart:559:55)
#11     MethodChannel.setMethodCallHandler.<anonymous closure> (package:flutter/src/services/platform_channel.dart:552:34)
#12     _DefaultBinaryMessenger.setMessageHandler.<anonymous closure> (package:flutter/src/services/binding.dart:567:35)
#13     _invoke2 (dart:ui/hooks.dart:344:13)
#14     _ChannelCallbackRecord.invoke (dart:ui/channel_buffers.dart:45:5)
#15     _Channel.push (dart:ui/channel_buffers.dart:135:31)
#16     ChannelBuffers.push (dart:ui/channel_buffers.dart:343:17)
#17     PlatformDispatcher._dispatchPlatformMessage (dart:ui/platform_dispatcher.dart:722:22)
[...]

PR #229, opened 2024-01-17
Use normalized text editing value for updating remote

Amir-P:
Fixes #227

PR #230, opened 2024-01-18
Fix down movement at end of document causing assertion errors

amantoux:
Fixes https://github.com/fleather-editor/fleather/issues/227

### DIFF
--- a/packages/fleather/lib/src/widgets/editor.dart
+++ b/packages/fleather/lib/src/widgets/editor.dart
@@ -2552,11 +2552,22 @@ class _UpdateTextSelectionVerticallyAction<
             : intent.forward
                 ? currentRun.moveNext()
                 : currentRun.movePrevious();
-    final TextPosition newExtent = shouldMove
-        ? currentRun.current
-        : intent.forward
-            ? TextPosition(offset: state.textEditingValue.text.length)
-            : const TextPosition(offset: 0);
+
+    TextPosition computeNewExtent() {
+      if (shouldMove) return currentRun.current;
+
+      if (intent.forward) {
+        if (collapseSelection) {
+          state.updateLastKnownWithSelection(TextSelection.collapsed(
+              offset: state.textEditingValue.text.length));
+        }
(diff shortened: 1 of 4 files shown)


Amir-P:
It looks good and we can use it too. But I'm thinking merging #229 is necessary even if it wasn't fixing the #227. When we're not rendering the last "\n", it shouldn't be sent and rendered by the underlying platforms too. WDYT? @amantoux

amantoux:
> uldn't be sent and rendered by the underlying platforms t

amantoux:
> It looks good and we can use it too. But I'm thinking merging #229 is necessary even if it wasn't fixing the #227. When we're not rendering the last "\n", it shouldn't be sent and rendered by the underlying platforms too. WDYT? @amantoux

Yes fine by me

packages/fleather/lib/src/widgets/editor.dart
```diff
@@ -2552,11 +2552,22 @@ class _UpdateTextSelectionVerticallyAction<
...
-        ? currentRun.current
-        : intent.forward
-            ? TextPosition(offset: state.textEditingValue.text.length)
-            : const TextPosition(offset: 0);
+
+    TextPosition computeNewExtent() {
+      if (shouldMove) return currentRun.current;
+
+      if (intent.forward) {
+        if (collapseSelection) {
```
Amir-P, line 2560: Why are we conditioning call to `updateLastKnownWithSelection` to `collapseSelection`? Wouldn't it cause the same problem in situations where selection is supposed to extend?

Amir-P (commented):
Thanks @amantoux ! Looks good to me overall but I have a question. PTAL.

packages/fleather/lib/src/widgets/editor.dart
amantoux: Because we don't want to set a collapses selection on remote if it is not.
In case of extended selection, there is no need for this I think

Amir-P (approved):

PR #233, opened 2024-01-20
Add and expose ClipboardManager

### LINKED ISSUE #56: Rich clipboard

Amir-P:
Closes #56.

Due to current clipboard limitations in Flutter, it's been decided not to provide a rich clipboard functionality out-of-box and instead keep using plain text clipboard functionality by default and provide a custom clipboard manager which can be implemented by user to set/read clipboard data in their own preferred way. This can change in future if Flutter starts to support enriched `ClipboardData`.


packages/fleather/lib/src/widgets/editor.dart
```diff
@@ -1166,13 +1196,25 @@ class RawEditorState extends EditorState
...
-    final ClipboardData? data = await Clipboard.getData(Clipboard.kTextPlain);
-    if (data == null) {
+    final data = await widget.clipboardManager.getData();
+    if (data?.plainText == null && data?.delta == null) {
       return;
     }
 
-    _replaceText(
-        ReplaceTextIntent(textEditingValue, data.text!, selection, cause));
+    if (data!.delta != null) {
```
amantoux, line 1204: may add `hasDelta` to `FleatherClipboardData`? (`hasHtml` also?)

packages/fleather/lib/src/widgets/editor.dart
```diff
@@ -1166,13 +1196,25 @@ class RawEditorState extends EditorState
     }
     // Snapshot the input before using `await`.
     // See https://github.com/flutter/flutter/issues/11427
-    final ClipboardData? data = await Clipboard.getData(Clipboard.kTextPlain);
-    if (data == null) {
+    final data = await widget.clipboardManager.getData();
+    if (data?.plainText == null && data?.delta == null) {
```
amantoux, line 1200: have `hasPlainText` and `hasDelta` in `FleatherClipboardData`

packages/fleather/lib/src/services/clipboard_manager.dart
```diff
@@ -0,0 +1,74 @@
...
+import 'package:parchment_delta/parchment_delta.dart';
+
+/// Encapsulates clipboard data
+///
+/// One of the properties should be non-null or null should be returned
+/// from [FleatherCustomClipboardGetData].
+/// When pasting data in editor, [delta] has precedence over [plainText].
+class FleatherClipboardData {
+  final String? plainText;
+  final Delta? delta;
```
amantoux, line 11: Don't we want to had possibility to paste HTML?

amantoux (changes requested):
Thanks @Amir-P ! It is a great feature!
I have some comments; PTAL

I also have one general comment: can we add in example an implementation using an external clipboard manager (such as super_clipboard)?

packages/fleather/lib/src/services/clipboard_manager.dart
Amir-P: IMHO, we should never add anything other than delta and plainText. Having html will add the complexity of handling custom embeds (e.g. image, video, ...). Also in case of implementing rich clipboard in Fleather, we need to handle other clipboard data like images. Last but not least, the differences in platforms and how they represent data in clipboard will introduce more complexity.

I think we can easily avoid all of this by providing users with 'FleatherCustomClipboardManager' and a decent html codec in Parchment. WDYT?

packages/fleather/lib/src/widgets/editor.dart
Amir-P: Check https://github.com/fleather-editor/fleather/pull/233#discussion_r1464335956 for 'hasDelta'.

Amir-P:
> I also have one general comment: can we add in example an implementation using an external clipboard manager (such as super_clipboard)?

Sure. In fact I had it in mind to do after getting this merged.

packages/fleather/lib/src/widgets/editor.dart
```diff
@@ -1166,13 +1196,25 @@ class RawEditorState extends EditorState
     }
     // Snapshot the input before using `await`.
     // See https://github.com/flutter/flutter/issues/11427
-    final ClipboardData? data = await Clipboard.getData(Clipboard.kTextPlain);
-    if (data == null) {
+    final data = await widget.clipboardManager.getData();
+    if (data?.hasPlainText != true && data?.hasDelta != true) {
```
amantoux, line 1200: Sorry, after thinking this through...
Could we have `if (data == null || data!.hasNoData)` ? (we can remove `hasPlainText`)
WDYT?

amantoux (changes requested):
PTAL

PR #242, opened 2024-01-24
Use overlay entries for selectors in toolbar

### LINKED ISSUE #202: I would like to when I tap on color buttons or SelectHeadingButton for the keyboard to say active. I find it a bit annoying the keyboard dismissing everytime user taps to change colors or headings
Right now keyboard dismises when we tap on color or heading buttons. After user select the color or the heading the keyboard shows up again but wouldnt it be better if the keyboard stayed always active? I would like to have that option

amantoux:
Closes #202


Amir-P:
Thanks, Alan! I just made some changes to improve the way selection is handled. Hope you find it useful. @amantoux

Amir-P (commented):
Thanks, Alan! I just made some changes to improve the way selection is handled. Hope you find it useful. @amantoux

Issue #274, opened 2024-02-09
Responsive toolbar layout


jacquetc:
**Is your feature request related to a problem? Please describe.**
Take your example app, resize it to be smaller than the width of the toolbar. Right-most buttons are hidden.

**Describe the solution you'd like**
See below, I prefer the first solution.

**Describe alternatives you've considered**
Different solutions:
- when the container width is smaller than the toolbar width, one drop down menu appears at the right to hold hidden buttons.
- the buttons can flow below, like in a second, third toolbar
- enable horizontal scrolling with ScrollView and arrows left and right to scroll manually for non-touch devices. I think this would be the solution adopted by the developer choosing Fleather since you can wrap the Toolbar without touching its internal.

Aggie85:
To be honest, it would be nice if there was a layout property and its choices were: scroll, wrap, moreMenu, etc...

Since I stared my Flutter app a little over a year ago and learning along the way, I am learning to be adaptive and creative in ways I was never forced before while programming in Windows for 30+ years. 

Flutter's cross-platform ability is very unrestrictive and thinking outside the box as opposed to being constrained by it like a lot of other frameworks is refreshing.

So allowing more options would be better.

Thanks for everyone's work!

All the best,

Aggie85

jacquetc:
Hello,

I'm not a developer of Feather, just a curious tourist doing sand castles, so the following is only my opinion:

As a developer yourself, you already understand that the multiplicity of options impacts the maintainability. 

I think that the developers times can be better used to missing core components before creating a more complex front-end. Also, at this stage of development of Fleather, front end can be easily impacted by profound changes in the back-end. 

I'm not saying here to not do something. I personally like your idea of options. There is already a FleatherToolbar.basic() factory, there can be others like FleatherToolbar.scroll(), etc...

Actually, you made me thinking on the problem and I see different avenues to choose: 
- Minimal maintenance: offer only the button widgets corresponding to each function. No toolbar, let the developer using Fleather bake whatever toolbar he wants.
- Widgets + basic toolbar with one basic responsive design. If a developer isn't happy with stock toolbar, he can bake its own with the offered widgets.
- Higher-level, higher-maintenance toolbars. Several toolbar styles as discussed before.

What do you think ?

If you want to share more of your point of view, a little debate is occurring in this PR: #277

PR #277, opened 2024-02-09
Responsive toolbar (#274)

jacquetc:
Hello,

For fun, in relation to #274, here is a preliminary work on a toolbar with an overflow menu. 
image

Left to do:
- Separators must become horizontal in the menu
- The menu must close after clicking on an item
- The Normal/Header widget doesn't display its drop-down menu. Same with color.
- The Normal/Header widget is a bit big, maybe insert instead an more little version of it in the menu instead ? 
image

### DIFF
--- a/packages/fleather/lib/src/widgets/editor_toolbar.dart
+++ b/packages/fleather/lib/src/widgets/editor_toolbar.dart
@@ -562,6 +562,50 @@ class _ColorPaletteElement extends StatelessWidget {
   }
 }
 
+/// Toolbar button which allows to display a dropdown menu with the buttons hidden because of the lack of space.
+
+class OverflowMenuButton extends StatefulWidget {
+  final List<Widget> menuChildren;
+
+  const OverflowMenuButton({super.key, required this.menuChildren});
+
+  @override
+  State<OverflowMenuButton> createState() => _OverflowMenuButtonState();
+}
+
+class _OverflowMenuButtonState extends State<OverflowMenuButton> {
+  @override
+  Widget build(BuildContext context) {
+    final theme = Theme.of(context);
+
+    return MenuAnchor(
+      builder:
+          (BuildContext context, MenuController controller, Widget? child) {
+        return FLIconButton(
+          highlightElevation: 0,
+          hoverElevation: 0,
+          size: 32,
+          icon: Icon(
+            Icons.more_vert,
+            size: 18,
+            color: theme.iconTheme.color,
+          ),
+          fillColor: theme.canvasColor,
+          onPressed: () {
+            if (controller.isOpen) {
+              controller.close();
+            } else {
+              controller.open();
+            }
+          },
+        );
+      },
+      anchorTapClosesMenu: true,
+      menuChildren: widget.menuChildren,
+    );
+  }
+}
+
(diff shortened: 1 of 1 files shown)


jacquetc:
Let me know if the idea is sound for you before I put more work to fix the remaining issues.

Aggie85:
This would be GREAT!

All the best,

Aggie85

jacquetc:
I think too that it's a lacking feature. Yet, there are other possible ways to deal with an overflowing menu. 

Also, my change feels dirty to me. I'm leveraging the FleatherToolbar.basic() factory. We can create instead a separate FleatherToolbar::overflowing() factory and dedicate a State build() for it, keeping a basic (and easily debug-able) Stateful builder clean for "basic" Toolbar. Bottom line: Adding functionalities, not touching things already running fine.

Trying to see what is possible since I'm learning Dart & Flutter on the go (C++ & Rust background here).

Waiting now for @amantoux or @Amir-P insights since they are active leading contributors. I don't want to break their vision of Fleather.

jacquetc:
image
With horizontal dividers.

amantoux:
I wonder if we should not propose a radically different layout for small displays.

jacquetc:
A few ideas: 
blazor-toolbar-item-in-overflow
blazor-toolbar-scrolling-touch
Sans titre
blazor-multi-row
toolbar-responsive-md
DsxUq

Amir-P:
I'm more in favor of what [quilljs](https://quilljs.com) is doing on their website. It wouldn't break what we have on big screens and makes interaction easier on smaller screens compared to an overflow menu. @jacquetc @amantoux

jacquetc:
quilljs's style can eat a lot of vertical space on smartphones if there is too much widgets to display. I'm finding this one to be a good compromise, so it's eating the visual space on demand.
304118662-3bcc466b-a53f-4432-9ff1-1fc00a6c9bac

amantoux:
Google Docs has a dynamic layout for it toolbar on mobile (which I fail to understand the behvior :))
Word provides a minimal toolbar with the possibility to show an advanced version
The latter has my preference (screenshots below)
IMG_1400
IMG_1401

jacquetc:
I am testing it on my phone. I didn't know that it was possible to put something at the bottom. It's quite nice and can be out of the way if there is no focus on the text zone. It would be a nice solution for Android and IOs, yet less for desktop where an always-visible toolbar (or advanced context menu like Office Word) would be preferred.

amantoux:
> I am testing it on my phone. I didn't know that it was possible to put something at the bottom. It's quite nice and can be out of the way if there is no focus on the text zone. It would be a nice solution for Android and IOs, yet less for desktop where an always-visible toolbar (or advanced context menu like Office Word) would be preferred.

Yes we would implement entirely different layouts (& behaviors) for desktop and mobile

amantoux:
@jacquetc Are you planning to move forward on this? Otherwise we will close this PR

jacquetc:
Hello, 
Beside the little hack I already did, I am too busy with family life to explore further. You can close the PR. You can also merge what was done, which is better than no responsive toolbar at all (I think ?).
Regards,
Cyril

PR #284, opened 2024-02-21
Fix and simplify selection update after pasting

Amir-P:
This PR fixes issues with previous attempt to fix selection update problem after pasting. It also simplifies pasting logic by using `controller.compose` for both plain text and delta clipboard data.


amantoux (approved):
LGTM!

PR #329, opened 2024-04-26
Fix toolbar selector positioning

### LINKED ISSUE #262: HeadigStyle Positions
Hi.

When i use Toolbar bottom Sticky,
image

i can't see HeadingStyle like  that
스크린샷 2024-02-01 210843

could you fix when i use bottom sticky toolbar to show heading style?

### LINKED ISSUE #290: When toggling colour selection on Toolbar, assertion received
## Steps to Reproduce

1. I use the toolbar as an overlay itself, a bit like Notion. However, when I toggle text colour option, I get the following assertion:

```
flutter: 'package:flutter/src/widgets/overlay.dart': Failed assertion: line 168 pos 12: '_overlay != null': is not true.
flutter: 
#0      _AssertionError._doThrowNew (dart:core-patch/errors_patch.dart:51:61)
#1      _AssertionError._throwNew (dart:core-patch/errors_patch.dart:40:5)
#2      OverlayEntry.remove (package:flutter/src/widgets/overlay.dart:168:12)
#3      SelectorScopeState.pushSelector.<anonymous closure> (package:fleather/src/widgets/editor_toolbar.dart:1259:20)
#4      _RootZone.run (dart:async/zone.dart:1655:54)
#5      _FutureListener.handleWhenComplete (dart:async/future_impl.dart:205:18)
#6      Future._propagateToListeners.handleWhenCompleteCallback (dart:async/future_impl.dart:816:39)
#7      Future._propagateToListeners (dart:async/future_impl.dart:872:11)
#8      Future._propagateToListeners (dart:async/future_impl.dart:771:9)
#9      Future._completeWithValue (dart:async/future_impl.dart:647:5)
#10     Future._asyncCompleteWithValue.<anonymous closure> (dart:async/future_impl.dart:721:7)
#11     _microtaskLoop (dart:async/schedule_microtask.dart:40:21)
#12     _startMicrotaskLoop (dart:async/schedule_microtask.dart:49:5)

flutter: ----------------------------------------------------
```
The colour selection menu does not open. 

## Environment
- MacOS
- Flutter version 3.16.4
- Fleather version 1.14.1

### LINKED ISSUE #317: It seems the new overlay usage in heading and color picking etc doesnt work when keyboard is active
Is this a known bug? when keyboard is active the overlay throwing an exception when i tap on headings or color picker

### LINKED ISSUE #323: Can anyone help me with overlay?
The color dialog show in wrong position:
WhatsApp Image 2024-04-13 at 10 32 39

```
import 'package:fleather/fleather.dart';
import 'package:flutter/material.dart';
import 'package:go_router/go_router.dart';

void main() {
  runApp(const MyApp());
}

class MyApp extends StatefulWidget {
  const MyApp({super.key});

  @override
  State<MyApp> createState() => _MyAppState();
}

class _MyAppState extends State<MyApp> {
  FleatherController controller =
      FleatherController(document: ParchmentDocument());

  @override
    return MaterialApp.router(
      routerConfig: GoRouter(
        initialLocation: '/',
        routes: <RouteBase>[
          ShellRoute(
            builder: (_, __, child) {
              return Scaffold(
                backgroundColor: Colors.grey,
                body: Container(
                  color: Colors.white,
                  margin: const EdgeInsets.only(top: 60, left: 200),
                  child: child,
                ),
              );
            },
            routes: [
              GoRoute(
                path: '/',
                builder: (_, __) => Column(
                  children: [
                    FleatherToolbar.basic(controller: controller),
                    Expanded(child: FleatherEditor(controller: controller)),
                  ],
                ),
              ),
            ],
          ),
        ],
      ),
    );
  }
}

```

Amir-P:
- Fixes #317 
- Fixes #290 
- Fixes #323
- Fixes #262 

- [x] Add tests
- [ ] Add breaking change migration guide

### DIFF
--- a/packages/fleather/lib/src/widgets/editor_toolbar.dart
+++ b/packages/fleather/lib/src/widgets/editor_toolbar.dart
@@ -419,7 +419,7 @@ class _ColorButtonState extends State<ColorButton> {
               onSelectedColor: completer.complete)),
     );
 
-    return SelectorScope.pushSelector(context, selector, completer);
+    return SelectorScope.showSelector(context, selector, completer);
   }
 
   @override
@@ -646,7 +646,7 @@ class _SelectHeadingButtonState extends State<SelectHeadingButton> {
       child: _HeadingList(theme: themeData, onSelected: completer.complete),
     );
 
-    return SelectorScope.pushSelector(context, selector, completer);
+    return SelectorScope.showSelector(context, selector, completer);
   }
 }
 
@@ -1208,11 +1208,11 @@ class SelectorScope extends StatefulWidget {
   static SelectorScopeState of(BuildContext context) =>
       context.findAncestorStateOfType<SelectorScopeState>()!;
 
-  static Future<T?> pushSelector<T>(
+  static Future<T?> showSelector<T>(
           BuildContext context, Widget selector, Completer<T?> completer,
           {bool rootOverlay = false}) =>
       SelectorScope.of(context)
-          .pushSelector(context, selector, completer, rootOverlay: rootOverlay);
+          .showSelector(context, selector, completer, rootOverlay: rootOverlay);
 
   @override
   State<SelectorScope> createState() => SelectorScopeState();
@@ -1221,23 +1221,23 @@ class SelectorScope extends StatefulWidget {
 class SelectorScopeState extends State<SelectorScope> {
   OverlayEntry? _overlayEntry;
 
-  /// The [context] should belong to the presentor widget.
-  Future<T?> pushSelector<T>(
+  /// The [context] should belong to the presenter widget.
+  Future<T?> showSelector<T>(
       BuildContext context, Widget selector, Completer<T?> completer,
       {bool rootOverlay = false}) {
     _overlayEntry?.remove();
 
     final overlay = Overlay.of(context, rootOverlay: rootOverlay);
 
-    final RenderBox presentor = context.findRenderObject() as RenderBox;
+    final RenderBox presenter = context.findRenderObject() as RenderBox;
     final RenderBox overlayBox =
         overlay.context.findRenderObject() as RenderBox;
-    final offset = Offset(0.0, presentor.size.height);
+    final offset = Offset(0.0, presenter.size.height);
     final position = RelativeRect.fromSize(
       Rect.fromPoints(
-        presentor.localToGlobal(offset, ancestor: overlayBox),
-        presentor.localToGlobal(
-          presentor.size.bottomRight(Offset.zero) + offset,
+        presenter.localToGlobal(offset, ancestor: overlayBox),
+        presenter.localToGlobal(
+          presenter.size.bottomRight(Offset.zero) + offset,
           ancestor: overlayBox,
         ),
       ),
@@ -1268,11 +1268,7 @@ class SelectorScopeState extends State<SelectorScope> {
         completer.complete(null);
       }
     });
-    completer.future.whenComplete(() {
-      _overlayEntry?.remove();
-      _overlayEntry?.dispose();
-      _overlayEntry = null;
-    });
+    completer.future.whenComplete(removeEntry);
     overlay.insert(_overlayEntry!);
     return completer.future;
   }
@@ -1280,13 +1276,8 @@ class SelectorScopeState extends State<SelectorScope> {
   void removeEntry() {
     if (_overlayEntry == null) return;
     _overlayEntry!.remove();
+    _overlayEntry!.dispose();
     _overlayEntry = null;
-  }
-
-  @override
-  void didUpdateWidget(covariant SelectorScope oldWidget) {
-    super.didUpdateWidget(oldWidget);
-    removeEntry();
   }
 
   @override
@@ -1347,21 +1338,13 @@ class _SelectorLayout extends SingleChildLayoutDelegate {
 
     // Find the ideal horizontal position.
     double x;
-    if (position.left > position.right) {
-      // Menu button is closer to the right edge, so grow to the left, aligned to the right edge.
-      x = size.width - position.right - childSize.width;
-    } else if (position.left < position.right) {
-      // Menu button is closer to the left edge, so grow to the right, aligned to the left edge.
-      x = position.left;
-    } else {
-      // Menu button is equidistant from both edges, so grow in reading direction.
-      switch (textDirection) {
-        case TextDirection.rtl:
-          x = size.width - position.right - childSize.width;
-        case TextDirection.ltr:
-          x = position.left;
-      }
+    switch (textDirection) {
+      case TextDirection.rtl:
+        x = size.width - position.right - childSize.width;
+      case TextDirection.ltr:
+        x = position.left;
     }
+    
     final Offset wantedPosition = Offset(x, y);
     final Offset originCenter = position.toRect(Offset.zero & size).center;
     final Iterable<Rect> subScreens =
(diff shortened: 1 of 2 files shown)


packages/fleather/lib/src/widgets/editor_toolbar.dart
```diff
@@ -1289,3 +1300,117 @@
...
+  Offset getPositionForChild(Size size, Size childSize) {
+    // size: The size of the overlay.
+    // childSize: The size of the menu, when fully open, as determined by
+    // getConstraintsForChild.
+
+    final double y = position.top;
+
+    // Find the ideal horizontal position.
+    double x;
+    if (position.left > position.right) {
```
amantoux, line 1350: Can we implement something that depends on TextDirection? 
We want to align to right if enough space in ltr and the other way round in rtl

packages/fleather/lib/src/widgets/editor_toolbar.dart
```diff
@@ -1242,27 +1221,59 @@ class SelectorScope extends StatefulWidget {
 class SelectorScopeState extends State<SelectorScope> {
   OverlayEntry? _overlayEntry;
 
-  Future<T?> pushSelector<T>(Widget selector, Completer<T?> completer) {
+  /// The [context] should belong to the presentor widget.
+  Future<T?> pushSelector<T>(
+      BuildContext context, Widget selector, Completer<T?> completer,
+      {bool rootOverlay = false}) {
```
amantoux, line 1227: We don't seem to use this option. We should remove it

packages/fleather/lib/src/widgets/editor_toolbar.dart
```diff
@@ -1228,12 +1201,18 @@ class FLIconButton extends StatelessWidget {
...
 
   static SelectorScopeState of(BuildContext context) =>
       context.findAncestorStateOfType<SelectorScopeState>()!;
 
-  final Widget child;
+  static Future<T?> pushSelector<T>(
+          BuildContext context, Widget selector, Completer<T?> completer,
+          {bool rootOverlay = false}) =>
+      SelectorScope.of(context)
+          .pushSelector(context, selector, completer, rootOverlay: rootOverlay);
```
amantoux, line 1215: We don't seem to use this option. We should remove it

packages/fleather/lib/src/widgets/editor_toolbar.dart
```diff
@@ -1242,27 +1221,59 @@ class SelectorScope extends StatefulWidget {
 class SelectorScopeState extends State<SelectorScope> {
   OverlayEntry? _overlayEntry;
 
-  Future<T?> pushSelector<T>(Widget selector, Completer<T?> completer) {
+  /// The [context] should belong to the presentor widget.
```
amantoux, line 1224: Typo = presenter?

packages/fleather/lib/src/widgets/editor_toolbar.dart
```diff
@@ -1242,27 +1221,59 @@ class SelectorScope extends StatefulWidget {
...
+      {bool rootOverlay = false}) {
     _overlayEntry?.remove();
-    final overlayEntry = OverlayEntry(
-      builder: (context) => TapRegion(
-        child: selector,
-        onTapOutside: (_) => completer.complete(null),
+
+    final overlay = Overlay.of(context, rootOverlay: rootOverlay);
+
+    final RenderBox presentor = context.findRenderObject() as RenderBox;
```
amantoux, line 1232: type=presenter?

amantoux (changes requested):
Many thank for this great enhancement!
I have some suggestions

packages/fleather/lib/src/widgets/editor_toolbar.dart
Amir-P: It's a public API and might be used by users implementing their own custom buttons. @amantoux

packages/fleather/lib/src/widgets/editor_toolbar.dart
Amir-P: It's a public API and might be used by users implementing their own custom buttons. @amantoux

packages/fleather/lib/src/widgets/editor_toolbar.dart
Amir-P: That's what I was thinking at first but thought it might be better to follow what Flutter's doing. But will change it since it's easier.

Amir-P:
Thanks for the review! I removed the removal of overlay entry on `didUpdateWidget` (causing a problem for one of the users #290) and applied your suggestions. Could you please take a look again? @amantoux

packages/fleather/lib/src/widgets/editor_toolbar.dart
amantoux: Ok, but still I think we shouldn't offer the option if not required by a user. It adds complexity for no assessed benefit.
You can keep it if you want though, I don't really mind

packages/fleather/lib/src/widgets/editor_toolbar.dart
amantoux: Ok, but still I think we shouldn't offer the option if not required by a user. It adds complexity for no assessed benefit.
You can keep it if you want though, I don't really mind

amantoux (changes requested):
PTAL

packages/fleather/lib/src/widgets/editor_toolbar.dart
Amir-P: Could you please provide the complete code for me to test the behavior?

amantoux:
```dart
@override
Offset getPositionForChild(Size size, Size childSize) {
  // size: The size of the overlay.
  // childSize: The size of the menu, when fully open, as determined by
  // getConstraintsForChild.

  final double y = position.top;

  // Find the ideal horizontal position.
  double x;
  if ((textDirection == TextDirection.ltr ? position.right : position.left) > childSize.width) {
    // Menu button is closer to the left edge, so grow to the right, aligned to the left edge.
    x = position.left;
  } else if ((textDirection == TextDirection.ltr ? position.left : position.right) > childSize.width) {
    // Menu button is closer to the right edge, so grow to the left, aligned to the right edge.
    x = size.width - position.right - childSize.width;
  } else {
    switch (textDirection) {
      case TextDirection.rtl:
        x = size.width - position.right - childSize.width;
      case TextDirection.ltr:
        x = position.left;
    }
  }

  final Offset wantedPosition = Offset(x, y);
  final Offset originCenter = position.toRect(Offset.zero & size).center;
  final Iterable<Rect> subScreens =
      DisplayFeatureSubScreen.subScreensInBounds(
          Offset.zero & size, avoidBounds);
  final Rect subScreen = _closestScreen(subScreens, originCenter);
  return _fitInsideScreen(subScreen, childSize, wantedPosition);
}
```

Amir-P: It's not working correctly for RTL case but I get the idea. I don't know which one is better. What do you say? @amantoux

Amir-P: In fact, it's not even working correctly for RTL right now. :))

amantoux:
> Thanks for the review! I removed the removal of overlay entry on `didUpdateWidget` (causing a problem for one of the users #290) and applied your suggestions. Could you please take a look again? @amantoux

Played with the new behavior, I think it's better than the old one

packages/fleather/lib/src/widgets/editor_toolbar.dart
amantoux: Difficult for me to test RTL, how bad is it?

Amir-P: I think I fixed it. Thanks for the suggestion! @amantoux

amantoux (approved):
LGTM!

PR #337, opened 2024-04-30
Add hard edge clip behavior to Material link bottom sheet modal

### LINKED ISSUE #331: Link menu lacks bottom padding and clipping
## Steps to Reproduce

1. Insert a link
2. Open the link menu
3. Use the buttons

The button at the bottom doesn't take into account the device controls and displays under them, some padding should be added.

The button at the top isn't clipped and displays outside the bottom sheet (when selected).

https://github.com/fleather-editor/fleather/assets/44336112/46f79758-cbb3-47a1-8458-60536a3c587f

## Environment
OS: Android
Flutter version: 3.19.3
Fleather version: 1.14.4

amantoux:
Fixes #331

### DIFF
--- a/packages/fleather/lib/src/widgets/link.dart
+++ b/packages/fleather/lib/src/widgets/link.dart
@@ -72,6 +72,7 @@ class _CupertinoAction extends StatelessWidget {
   final String title;
   final IconData icon;
   final VoidCallback onPressed;
+
   const _CupertinoAction({
     required this.title,
     required this.icon,
@@ -109,6 +110,7 @@ class _CupertinoAction extends StatelessWidget {
 Future<LinkMenuAction> _showMaterialMenu(
     BuildContext context, String link) async {
   final result = await showModalBottomSheet<LinkMenuAction>(
+    clipBehavior: Clip.hardEdge,
     context: context,
     builder: (ctx) {
       return Column(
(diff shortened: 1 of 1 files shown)


Amir-P (commented):
Happy to see you back again!
I think the OP had an issue with safe are and bottom padding too. Could you please fix that too?

amantoux:
> Happy to see you back again! I think the OP had an issue with safe are and bottom padding too. Could you please fix that too?

Not sure why the padding would be needed
I added `useSafeArea`. (Funny the default is `false`)

packages/fleather/lib/src/widgets/link.dart
```diff
@@ -109,6 +110,8 @@ class _CupertinoAction extends StatelessWidget {
 Future<LinkMenuAction> _showMaterialMenu(
     BuildContext context, String link) async {
   final result = await showModalBottomSheet<LinkMenuAction>(
+    clipBehavior: Clip.hardEdge,
+    useSafeArea: true,
```
Amir-P, line 114: As it's been pointed out in https://github.com/fleather-editor/fleather/issues/331#issuecomment-2088701131, `useSafeArea` does not fix the issue. I tried wrapping the widget inside `builder` with `SafeArea` and seems to be working fine. Could you check it?

Amir-P (changes requested):
PTAL.

packages/fleather/lib/src/widgets/link.dart
amantoux: As discussed, @Amir-P noticed the `useSafeArea` argument does not include padding with respect to the bottom.

Amir-P (approved):
LGTM! Thanks!

PR #339, opened 2024-04-30
List numbers should follow the theme

### LINKED ISSUE #264: Adjusting TextBlockTheme height does not reflect in block type.
## Steps to Reproduce

1. Adjust a TextBlockTheme. E.g.:
```
  lists: TextBlockTheme(style: const TextStyle(
                                  fontWeight: FontWeight.w400,
                                  height: 1.55,
                                  fontSize: 14
                              ), spacing: VerticalSpacing(

                              ), ),
```
2. See how a numbered list does not align with text. 

Screenshot 2024-02-03 at 21 50 29

## Environment
- OS: Mac OS
- Flutter version 3.16.4
- Fleather version 1.13.2

amantoux:
Closes #264

### DIFF
--- a/packages/fleather/lib/src/widgets/editable_text_block.dart
+++ b/packages/fleather/lib/src/widgets/editable_text_block.dart
@@ -163,7 +163,7 @@ class EditableTextBlock extends StatelessWidget {
 
       leadingWidgets.add(_NumberPoint(
         number: currentIndex + 1,
-        style: theme.paragraph.style,
+        style: theme.lists.style,
         width: 32.0,
         padding: 8.0,
       ));
@@ -318,8 +318,13 @@ class _NumberPoint extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    final styleHeight = style.height;
     return Container(
-      alignment: AlignmentDirectional.topEnd,
+      // Empirically, depending on height of the style, we need to
+      // align at top or at bottom
+      alignment: styleHeight == null || styleHeight <= 1
+          ? AlignmentDirectional.topEnd
+          : AlignmentDirectional.bottomEnd,
       width: width,
       padding: EdgeInsetsDirectional.only(end: padding),
       child: Text(withDot ? '$number.' : '$number', style: style),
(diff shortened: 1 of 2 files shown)


amantoux:
@Amir-P no tests here neither, sounds like overkill. Ok with you?

packages/fleather/lib/src/widgets/editable_text_block.dart
```diff
@@ -318,8 +318,13 @@ class _NumberPoint extends StatelessWidget {
 
   @override
   Widget build(BuildContext context) {
+    final styleHeight = style.height;
     return Container(
-      alignment: AlignmentDirectional.topEnd,
+      // Empirically, depending on height of the style, we need to
+      // align at top or at bottom
+      alignment: styleHeight == null || styleHeight <= 1
```
amantoux, line 325: @Amir-P If you can find anything better, it'll be most welcome
I'm not satisfied with this to be honest

Amir-P (commented):
Thanks for the PR. The bottom alignment seems to fail when the text inside list breaks into multiple lines. In the screenshot, the top editor uses 1.8, the one in the middle 1.3 and the bottom one uses 1.0 as style height.

Screenshot 2024-05-05 at 11 41 26

amantoux:
> Thanks for the PR. The bottom alignment seems to fail when the text inside list breaks into multiple lines. In the screenshot, the top editor uses 1.8, the one in the middle 1.3 and the bottom one uses 1.0 as style height.
> Screenshot 2024-05-05 at 11 41 26
> 
> I think we need to align the leading widget to the baseline instead of top or bottom. What do you think? Here is the same styles for unordered list:
> Screenshot 2024-05-05 at 11 45 26

Yes aligning to baseline is way to go.

Amir-P:
> @Amir-P shouldn't we address this in a separate PR?

You mean merging it as is? I don't find it in a mergeable state since it's introducing a new issue which is more severe than the one it's fixing. @amantoux

amantoux:
> > @Amir-P shouldn't we address this in a separate PR?
> 
> You mean merging this one as is? I don't find it in a mergeable state since it's introducing a new issue which is more severe than the one it's fixing. @amantoux

No obviously
Restoring alignment

Amir-P (approved):
LGTM. We will address the alignment issue in another PR.

PR #345, opened 2024-05-08
Add checkbox support to markdown codec

### LINKED ISSUE #344: Markown conversion should support checkboxes
## Steps to Reproduce

1. Add checkboxes
2. Convert to markdown (with `ParchmentMarkdownCodec().encode(document)`)

## Environment
- OS: Android
- Flutter version: 3.19.6
- Fleather version: 1.14.5+1

Export to Markdown doesn't support every formatting that the editor does, and that is OK as alignment or underline or not supported by Markdown in the first place. However, checkboxes are.

Empty checkboxes are currently converted to "null", and checked ones produce an error (`Invalid argument(s): Cannot handle checked: true`).

Furthermore, maybe a function `canEncode` could be useful to know if a document can be converted, instead of needing to catch an exception when it is not?

Amir-P:
Fixes #344

- [X] Encoder
- [x] Decoder


amantoux (approved):
LGTM

PR #347, opened 2024-05-09
HTML codec: Support indentation decoding

### LINKED ISSUE #295: ParchmentHtmlCodec().decode can't convert indent

amantoux:
Only one level supported

Closes #295


Amir-P (approved):
Thanks! LGTM for now. Just wondering how we can adapt it for RTL cases?

amantoux:
> wondering how we can adapt it for RTL cases?

You mean, if we can detect that text direction is rtl and use padding-right instead of left?

Amir-P:
> > wondering how we can adapt it for RTL cases?
> 
> You mean, if we can detect that text direction is rtl and use padding-right instead of left?

Honestly, I have no idea right now. That might not be enough since the language of text doesn't determine the directionality of elements.

PR #351, opened 2024-05-11
Remove selection gestures when enableInteractiveSelection is `false`

### LINKED ISSUE #332: Read-only mode consumes gestures
I have set FleatherEditor to read-only mode for rendering content, but it interferes with the PageView component.

In the end, I can only disable some gestures, but it's still not ideal because I have custom components within Fleather that rely on horizontal swipe gestures. Is there a way to disable the gestures of FleatherEditor without affecting the gestures of the internal components?

https://github.com/fleather-editor/fleather/assets/79079747/51eda50d-6a20-4004-9ce9-2b52c43b11ae

amantoux:
Closes #332 

- [x] add tests

### DIFF
--- a/packages/fleather/lib/src/widgets/editor.dart
+++ b/packages/fleather/lib/src/widgets/editor.dart
@@ -491,10 +491,10 @@ class _FleatherEditorState extends State<FleatherEditor>
       ),
     );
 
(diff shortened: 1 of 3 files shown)


Amir-P (approved):
LGTM.

PR #359, opened 2024-05-29
Enable auto correct

simonbengtsson:
I thought it would be a major undertaking, but after some research it turns out toggling the autocorrect flag was enough. Is there a reason this was turned off or can it be enabled? Auto correct was one of the main reasons that made me still reach for Apple Notes at times.

Note that I'm using Google Keyboard in the attached demo and auto correct also worked when I tried with SwiftKey. I could not get it to work with the default iOS keyboard (not in English nor Swedish). Any thoughts on that? My guess it is some kind of locale setting either on my device or with the Fleather editor somehow.

https://github.com/fleather-editor/fleather/assets/3586691/efdff54c-34a7-489b-9a6f-68955c456495

### DIFF
--- a/packages/fleather/lib/src/widgets/editor.dart
+++ b/packages/fleather/lib/src/widgets/editor.dart
@@ -155,6 +155,11 @@ class FleatherEditor extends StatefulWidget {
   /// Defaults to `false`. Must not be `null`.
   final bool readOnly;
 
+  /// Whether to enable autocorrection.
+  ///
+  /// Defaults to `true`.
+  final bool autocorrect;
+
   /// Whether to enable user interface affordances for changing the
   /// text selection.
   ///
@@ -283,6 +288,7 @@ class FleatherEditor extends StatefulWidget {
     this.autofocus = false,
     this.showCursor = true,
     this.readOnly = false,
+    this.autocorrect = true,
     this.enableInteractiveSelection = true,
     this.minHeight,
     this.maxHeight,
@@ -551,6 +557,7 @@ class RawEditor extends StatefulWidget {
     this.autofocus = false,
     bool? showCursor,
     this.readOnly = false,
+    this.autocorrect = true,
     this.enableInteractiveSelection = true,
     this.minHeight,
     this.maxHeight,
@@ -601,6 +608,11 @@ class RawEditor extends StatefulWidget {
   ///
   /// Defaults to false. Must not be null.
   final bool readOnly;
+
+  /// Whether to enable autocorrection.
+  ///
+  /// Defaults to `true`.
+  final bool autocorrect;
 
   /// Callback which is triggered when the user wants to open a URL from
   /// a link in the document.
--- a/packages/fleather/lib/src/widgets/editor.dart
+++ b/packages/fleather/lib/src/widgets/editor.dart
@@ -457,6 +457,7 @@ class _FleatherEditorState extends State<FleatherEditor>
       scrollable: widget.scrollable,
       padding: widget.padding,
       autofocus: widget.autofocus,
+      autocorrect: widget.autocorrect,
       showCursor: widget.showCursor,
       readOnly: widget.readOnly,
       enableInteractiveSelection: widget.enableInteractiveSelection,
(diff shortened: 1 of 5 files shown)


Amir-P (changes requested):
Thank you for contributing! I personally disable the auto correction the moment I buy my phones so never noticed we're missing auto correction. But it would make sense to have a flag to control it from 'FleatherEditor' and 'FleatherField'.

simonbengtsson:
Oh really? You must have amazing finger control. Updated PR now with a new `autocorrect` flag for both `FleatherEditor` and `FleatherField`. Something like this you had in mind? Used the same doc comment used for the `TextField` widget in flutter.

Amir-P (commented):
Thanks! I left a comment for you. PTAL.

packages/fleather/lib/src/widgets/editor.dart
```diff
@@ -602,6 +609,11 @@ class RawEditor extends StatefulWidget {
   /// Defaults to false. Must not be null.
   final bool readOnly;
 
+  /// Whether to enable autocorrection.
+  ///
+  /// Defaults to `true`.
+  final bool autocorrect;
```
Amir-P, line 615: The `FleatherEditor` needs to pass the `autocorrect` to `RawEditor`.

packages/fleather/lib/src/widgets/editor_input_client_mixin.dart
```diff
@@ -57,7 +57,7 @@ mixin RawEditorStateTextInputClientMixin on EditorState
           inputType: TextInputType.multiline,
           readOnly: widget.readOnly,
           obscureText: false,
-          autocorrect: false,
+          autocorrect: widget.autocorrect,
```
Amir-P, line 60: Could you please see if you can add a test for this? You can check https://github.com/fleather-editor/fleather/blob/34b56e7f69cb76f5dfaf1b40219c010a50aa638d/packages/fleather/test/widgets/editor_input_client_mixin_test.dart#L11-L24 for writing the test.

Amir-P (changes requested):
I accidentally deleted my previous review. Left you some comments please take a look.

Amir-P:
> Note that I'm using Google Keyboard in the attached demo and auto correct also worked when I tried with SwiftKey. I could not get it to work with the default iOS keyboard (not in English nor Swedish). Any thoughts on that? My guess it is some kind of locale setting either on my device or with the Fleather editor somehow.

BTW, the auto correction seems to be working fine on iOS too.

https://github.com/fleather-editor/fleather/assets/8766492/0b4c5ab5-79ef-4f84-9ea2-4178a19c8a2f

packages/fleather/lib/src/widgets/editor.dart
simonbengtsson: Oops! Fixed now.

packages/fleather/lib/src/widgets/editor_input_client_mixin.dart
simonbengtsson: Did you have in mind a test something like this?

```dart
testWidgets('enters text that should be auto corrected', (tester) async {
  bind(tester);
  final document = ParchmentDocument();
  final editor = EditorSandBox(tester: tester, document: document);
  await editor.pump();
  await editor.tap();
  tester.binding.scheduleWarmUpFrame();
  await tester.pumpAndSettle();
  await editor.enterText(const TextEditingValue(text: 'hello wolrd '));
  await tester.pumpAndSettle(throttleDuration);
  final text = editor.document.toPlainText();
  expect(
    text,
    'hello world',
  );
});
```

I have yet to find a find a way to mock the auto correction service, but will research it further. I would have expected this to work, but it is never called.

```dart
void bind(WidgetTester tester) {
   // ...
  tester.binding.defaultBinaryMessenger.setMockMethodCallHandler(
      SystemChannels.spellCheck, (MethodCall methodCall) async {
    print("SystemChannels.spellCheck ${methodCall.method}");
    return null;
  });
}
```

Amir-P: Just testing that we are enabling/disabling the autocorrect when creating `TextInputConnection` would be enough. I believe you can check for `TextInput.setClient` method call using `setMockMethodCallHandler`. If you want, I can push a commit for the test.

simonbengtsson: Make sense! Added now. Looks good?

Amir-P (approved):
Thanks, LGTM! @amantoux Do you wanna take a look too?

amantoux (approved):
LGTM

PR #364, opened 2024-06-09
Add enable suggestions

Amir-P:
The `enableInteractiveSelection` is not tested because editor does not catch taps when it's `false`.


amantoux (approved):
LGTM!

PR #365, opened 2024-06-09
Fix theming for dark themes

### LINKED ISSUE #363: Improve the colors of inline code and code blocks
## Steps to Reproduce

1. Add some inline code or a code block

I find the colors of inline code and code blocks make them hard to use:
- In light mode (pure white background), the code block especially is so light that I can barely distinguish it from the background. I prefer the background color of the inline code.
- In dark mode, and even more in black mode, the background of the inline code and the code block should really be a dark gray. The inline code also has some really bad contrast that make it hard to read.

## Environment
- OS: Android
- Flutter version: 3.22
- Fleather version: 1.14.5+1

maelchiotti:
Fixes #363.

Changes:

- Horizontal rule: all theme
- Inline code: radius (to be the same as code blocks), background color
- Quote: background color
- Code block: text color and opacity (to be the same as inline code), background color

### DIFF
--- a/packages/fleather/lib/src/widgets/editor.dart
+++ b/packages/fleather/lib/src/widgets/editor.dart
@@ -84,8 +84,7 @@ Widget defaultFleatherEmbedBuilder(BuildContext context, EmbedNode node) {
     final fleatherThemeData = FleatherTheme.of(context)!;
 
     return Divider(
-      height: fleatherThemeData.paragraph.style.fontSize! *
-          fleatherThemeData.paragraph.style.height!,
+      height: fleatherThemeData.horizontalRule.height,
       thickness: fleatherThemeData.horizontalRule.thickness,
       color: fleatherThemeData.horizontalRule.color,
     );
--- a/packages/fleather/lib/src/widgets/editor.dart
+++ b/packages/fleather/lib/src/widgets/editor.dart
@@ -81,14 +81,13 @@ typedef FleatherEmbedBuilder = Widget Function(
 /// Only supports "horizontal rule" embeds.
 Widget defaultFleatherEmbedBuilder(BuildContext context, EmbedNode node) {
   if (node.value.type == 'hr') {
-    final theme = Theme.of(context);
-    final fleatherTheme = FleatherTheme.of(context)!;
+    final fleatherThemeData = FleatherTheme.of(context)!;
 
     return Divider(
-      height: fleatherTheme.paragraph.style.fontSize! *
-          fleatherTheme.paragraph.style.height!,
-      thickness: 2,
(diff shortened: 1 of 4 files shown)


packages/fleather/lib/src/widgets/editor.dart
```diff
@@ -81,11 +81,14 @@ typedef FleatherEmbedBuilder = Widget Function(
...
+    final theme = Theme.of(context);
+    final fleatherTheme = FleatherTheme.of(context)!;
+
     return Divider(
-      height: theme.paragraph.style.fontSize! * theme.paragraph.style.height!,
+      height: fleatherTheme.paragraph.style.fontSize! *
+          fleatherTheme.paragraph.style.height!,
       thickness: 2,
-      color: Colors.grey.shade200,
+      color: theme.colorScheme.surfaceContainerHigh,
```
Amir-P, line 91: I think it would be better to add something like `HorizontalRuleTheme` to `FleatherTheme` to control thickness and color.

Amir-P (changes requested):
Thank you for contributing. I really like the new colors. I just had a small concern with horizontal rule. Please take a look.

packages/fleather/lib/src/widgets/editor.dart
maelchiotti: Done!

packages/fleather/lib/src/widgets/editor.dart
```diff
@@ -81,11 +81,13 @@ typedef FleatherEmbedBuilder = Widget Function(
...
 Widget defaultFleatherEmbedBuilder(BuildContext context, EmbedNode node) {
   if (node.value.type == 'hr') {
-    final theme = FleatherTheme.of(context)!;
+    final fleatherThemeData = FleatherTheme.of(context)!;
+
     return Divider(
-      height: theme.paragraph.style.fontSize! * theme.paragraph.style.height!,
-      thickness: 2,
-      color: Colors.grey.shade200,
+      height: fleatherThemeData.paragraph.style.fontSize! *
```
Amir-P, line 87: Now it makes sense to read the height from the `HorizontalRuleTheme` too with `fleatherThemeData.paragraph.style.fontSize! * fleatherThemeData.paragraph.style.height!` as the default value. What do you think?

maelchiotti: Done

Amir-P (approved):
LGTM. Thanks again. @maelchiotti 
Do you want to take a look too? @amantoux

amantoux (approved):
LGTM! Thank you for the contribution!

PR #367, opened 2024-06-10
Support inline markdown shortcuts

### LINKED ISSUE #355: Can you add support for bold, strikethrough, underline, and italic Markdown syntax?

Amir-P:
Fixes #355 by adding markdown shortcuts for bold, italic, inline code and strikethrough. Looks like underline isn't a standard attribute in [markdown](https://www.markdownguide.org/cheat-sheet/).

- [ ] Add tests


amantoux (approved):
LGTM

PR #370, opened 2024-06-22
Release v1.16.0

amantoux:


packages/fleather/CHANGELOG.md
```diff
@@ -1,3 +1,11 @@
+## 1.16.0
+
+* Inline Markdown shortcuts
+* Enable auto-correct on mobile devices
+* Enable suggestions on Android devices
+* Adjust theme to dark mode
```
Amir-P, line 6:
```suggestion
* [Behavior change] Add auto-correct which defaults to `true`
* [Behavior change] Add enableSuggestions which defaults to `true`
* [UI change] Adjust default theme
```

Amir-P (commented):
Thanks! Just left you a change suggestion for `CHANGELOG.md` to indicate the breaking changes.

packages/fleather/CHANGELOG.md
amantoux: I don't think we should add to many tags to the release note
It is not a breaking change it is an enhancement
Nevertheless, we must precise that auto correct only works for mobile and suggestions only on Android

Amir-P: I still prefer to let the user know it's going to break the UI and behavior but your version is fine too. Just please change the theme changelog since it affects the light theme too.

Amir-P (approved):
LGTM. Thanks.

PR #372, opened 2024-06-25
Add Custom AutoFormats with Fallback Autoformats

cotw-fabier:
Hi Guys,

One more PR here based on https://github.com/fleather-editor/fleather/issues/369. This introduces a simple factory to the AutoFormat class which allows us to build it with the fallback. To keep things simple I opted to initalize the fallback and then merge that with my new list of AutoFormats. I know this isn't the most efficient but I didn't know if you wanted to leave the fallback() factory alone. Another option would have been to introduce a nullable variable so we could just add them there which would also be pretty ergonomic. 

If you prefer that I could roll that in and update the PR here.

I created a new example file with autoformats and also a sample for rendering based on the issue mentioned above. Simple, but should be advanced enough so people can understand the API without too much trouble!

Thanks for the awesome package!

### DIFF
--- a/packages/fleather/example/lib/autoformat_iterator.dart
+++ b/packages/fleather/example/lib/autoformat_iterator.dart
@@ -1,0 +1,364 @@
+import 'dart:convert';
+import 'dart:io';
+
+import 'package:fleather/fleather.dart';
+import 'package:flutter/foundation.dart';
+import 'package:flutter/material.dart';
+import 'package:flutter/services.dart';
+import 'package:image_picker/image_picker.dart';
+import 'package:parchment_delta/parchment_delta.dart';
+import 'package:url_launcher/url_launcher.dart';
+
+void main() {
+  runApp(const FleatherApp());
+}
+
+class FleatherApp extends StatelessWidget {
+  const FleatherApp({Key? key}) : super(key: key);
+
+  @override
+  Widget build(BuildContext context) => MaterialApp(
+        debugShowCheckedModeBanner: false,
+        theme: ThemeData.light(),
+        darkTheme: ThemeData.dark(),
+        title: 'Fleather - rich-text editor for Flutter',
+        home: HomePage(),
+      );
+}
+
+class HomePage extends StatefulWidget {
+  const HomePage({Key? key}) : super(key: key);
+
+  @override
+  _HomePageState createState() => _HomePageState();
+}
+
+class _HomePageState extends State<HomePage> {
+  final FocusNode _focusNode = FocusNode();
+  FleatherController? _controller;
+
+  @override
+  void initState() {
+    super.initState();
+    if (kIsWeb) BrowserContextMenu.disableContextMenu();
+    _initController();
+  }
+
+  @override
+  void dispose() {
+    super.dispose();
+    if (kIsWeb) BrowserContextMenu.enableContextMenu();
+  }
+
+  Future<void> _initController() async {
+    try {
+      final result = await rootBundle.loadString('assets/welcome.json');
+
+      /// Build Autoformats with backups
+      /// Autoformats allow for ergonomic automatic text transformations.
+      /// This example takes ![youtube link] and transforms it into a youtube blockembed.
+      /// Fallback text transformations apply styles by using markdown such as _italics_ or **bold**.
+      final customAutoFormat = AutoFormatYoutubeEmbed();
+      final autoFormats = AutoFormats.buildWithFallback([customAutoFormat]);
+
+      /// Heuristics work very similar to autoformats but are focused on improving the editing experience.
+      final heuristics = ParchmentHeuristics(
+        formatRules: [],
+        insertRules: [
+          ForceNewlineForInsertsAroundInlineImageRule(),
+        ],
+        deleteRules: [],
+      ).merge(ParchmentHeuristics.fallback);
+      final doc = ParchmentDocument.fromJson(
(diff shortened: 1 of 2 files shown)


packages/fleather/example/lib/autoformat_iterator.dart
```diff
@@ -0,0 +1,364 @@
+import 'dart:convert';
```
Amir-P, line 1: It might be better to have an example app showcasing every feature we have instead of a separate example for each feature, and a simple example to get users started quickly. I suggest to remove this example from this PR and create another PR for a complete advanced example. WDYT?

packages/fleather/lib/src/widgets/autoformats.dart
```diff
@@ -38,6 +38,18 @@ class AutoFormats {
     ]);
   }
 
+  /// Create a new instance of [AutoFormats] with the provided [autoFormats] + fallback [AutoFormats].
```
Amir-P, line 41: Could you please make current auto format implementations (`_AutoTextDirection`, `_MarkdownInlineShortcuts`, `_MarkdownLineShortcuts`) public so that users have the option of choosing from?

Amir-P (commented):
Thank you for contributing @cotw-fabier! Your changes on the `AutoFormats` looks good to me. Left you two comments. PTAL.

packages/fleather/example/lib/autoformat_iterator.dart
cotw-fabier: Hi @Amir-P,

This is really the main example with some additions to showcase this extra feature so it could overwrite the example app you have there already if you choose to use it. I have been taking notes with the intention to eventually help with some of the documentation and covering several use cases with Fleather and Parchment. Trying to build a few cohesive examples. This and https://github.com/fleather-editor/fleather/pull/371 both have examples working with the same "youtube" custom block example. 

However, I'm not married to any of it! Definitely take what you like and leave the rest. :)

packages/fleather/lib/src/widgets/autoformats.dart
cotw-fabier: Ok, Took a second swing here. Removed buildWithFallback and just extended the fallback method. I added a public enum and some params (autoFormats and defaultAutoFormats) so you can control how fallback builds the AutoFormats class. It will gracefully fall back to its original behavior if no params are provided, or if you just add your own params it'll add the fallback params automatically. But now you can also build a list of desired fallback formats and it will allow you to pick and choose as desired.

I left the merge function I originally had in there because it seems ergonomic for future use but its unused at this point so it could be removed.

This is the commit with my changes: https://github.com/fleather-editor/fleather/pull/372/commits/41e72fc255277faa2989d3a81e2be09f00f363cc

Then I realized I had left a spread in there from an earlier rendition of my changes so I made a second commit to just strip that out: https://github.com/fleather-editor/fleather/pull/372/commits/bf1fb43f8a684de7c2b74d8f1e04109a1ecaf4ac

Happy to make further changes as needed.

packages/fleather/example/lib/autoformat_iterator.dart
Amir-P: In that case let's rename this into something like `example_advanced.dart`, then we can simplify current example we have in another PR by removing inline images and embeds for users to get started with basics quicker. Also I can see you've defined `YoutubeEmbed` but never used it.

packages/fleather/lib/src/widgets/autoformats.dart
Amir-P: I'm gonna suggest a change for this if you don't mind.

packages/fleather/example/lib/autoformat_iterator.dart
cotw-fabier: Sounds good to me. I also realized I left the old version of the autoformats.buildWithFallback command in the example. Let me clean it up and post a new commit. Should be able to do it closer to the weekend -- need to wrap up a work project tomorrow.

packages/fleather/lib/src/widgets/autoformats.dart
cotw-fabier: Oh.... Thats fine with me. I would have to track down how they became private to begin with. I didn't write that bit. I'll check it out soon.

Amir-P: They were made private from the first commit they were added. Removing the underscore in their names would make them public.

Amir-P:
Hey @cotw-fabier. Are you still interested in moving this forward?

cotw-fabier:
> Hey @cotw-fabier. Are you still interested in moving this forward?

Amir,

Sorry this fell down my list a bit but I am still quite interested. I took a second to look at this and it seem to make the change you're asking for I would have to rename the functions being called along with changing their reference. I think this is probably fine, but wanted to confirm?

I'm going to be out of town for a few days but will revisit next week when I return.

Amir-P:
> Sorry this fell down my list a bit but I am still quite interested. I took a second to look at this and it seem to make the change you're asking for I would have to rename the functions being called along with changing their reference. I think this is probably fine, but wanted to confirm?

I'm not sure which functions you're talking about. What we discussed was to make the current auto formats public so that users can use a subset of them to instantiate `AutoFormats` class and also have `AutoFormats.fallback` to accept a list of additional auto formats. BTW, if it's still unclear to you, I can push the commit and you take care of example.

PR #375, opened 2024-07-05
Do not request keyboard on controller update

### LINKED ISSUE #301: Do not open the keyboard when a checkbox is toggled
## Steps to Reproduce

On mobile, without the keyboard opened, toggle a checkbox. The keyboard will open.

Not only this is not desired, but if the cursor is very far away from the checkbox, the editor will jump to it.

## Environment

- Android (and I guess iOS)
- Flutter version 3.19.3
- Fleather version 1.14.2

Amir-P:
Seems like removal of `requestKeyboard` on controller update was enough. Fixes #301.


amantoux (approved):
LGTM

PR #376, opened 2024-07-14
Ability to provide a custom clipboard status notifier

amantoux:
Current when providing a custom clipboard manager, the editor only shows _Paste_ in the toolbar if `text/plain` data is available.

A client might need to know if custom data (a delta serialization format) is available in clipboard, but it needs to be implemented by the client.

This PR also fix an issue where the _Paste_ would not appear in toolbar although data is available in clipboard.


Amir-P (approved):
LGTM.

PR #384, opened 2024-07-28
Add performance tests

Amir-P:


.github/workflows/fleather.yml
```diff
@@ -1,6 +1,10 @@
 name: build
 
-on: [push, pull_request]
```
amantoux, line 3: I find it interesting to have the CI being run on a the branch before creating the pull request

.github/workflows/performance_test.yaml
```diff
@@ -0,0 +1,48 @@
...
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - uses: subosito/flutter-action@v2
+        with:
+          flutter-version: "3.22.0"
+          cache: true
+          # Manually Update this `key`
+          cache-key: "20230512"
```
amantoux, line 21: We should update the cache key (keep on forgetting) if we want to benefit from the caching mechanism

amantoux (changes requested):
Thanks @Amir-P for putting this in place
It looks good to me.
I have a general concern: Aren't these tests run on a virtual machine? If it is the case we might observe volatility on the result and possibly false negatives/positives?
Any thoughts on this?

amantoux (approved):

.github/workflows/fleather.yml
Amir-P: The reason for limiting push to master branch was the duplicated run of CI on PR but I think using push as trigger would solve it and enable running CI on every branch without PR too. I'll test it.

.github/workflows/performance_test.yaml
Amir-P: Why is that?

amantoux: Otherwise it won't refresh the cache
https://github.com/subosito/flutter-action

Amir-P: Looks like using version and channel in the cache key might be better. I'll make the change. Let me know what you think.

Amir-P:
I talked to Alan outside of GitHub, and we decided to keep it as is but improve it in future for more streamlined local runs. We can also run on master and the current HEAD in the same virtual machine to have more realistic and stable performance tests (and maybe run multiple times and average the values).

.github/workflows/fleather.yml
```diff
@@ -1,10 +1,6 @@
 name: build
 
-on:
-  push:
-    branches:
-      - 'master'
-  pull_request:
+on: push
```
amantoux, line 3: I found this https://github.com/orgs/community/discussions/26276#discussioncomment-3251140
Isn't what we want?

amantoux (approved):
Left a comment, let me know your thought

.github/workflows/fleather.yml
Amir-P: I thought you wanted to have the CI run for other branches without a PR? What you've suggested wont do that.
Is there a problem with what we have now? It runs workflow for every push on any branches.

https://github.com/fleather-editor/fleather/pull/384#discussion_r1701325530

amantoux: No I think we should run at least on master and PRs
I'm fine with running it on branches and PRs

Amir-P: I think what I did solves the duplicated run and it would also run for other branches. @amantoux

.github/workflows/performance_test.yaml
amantoux: Yes also possible this way, it is just a label

.github/workflows/fleather.yml
amantoux: As discussed, we need to do some more research

amantoux (approved):
LGTM

packages/fleather/example/pubspec.yaml
```diff
@@ -15,6 +15,8 @@ dependencies:
   image_picker: ^1.0.7
   fleather:
     path: ../
+  parchment:
```
amantoux, line 18: Why do we need this?

amantoux (commented):
Thanks for putting this in place
I have a tiny comment

packages/fleather/example/pubspec.yaml
Amir-P: Analyzer started warning about usage of `import 'package:parchment/codecs.dart';` in `editing_test.dart`. @amantoux

PR #386, opened 2024-08-02
Upgrade example dependencies

amantoux:
Example doesn't compile on web due to obsolete web dependency
This fixes the issue


amantoux:
`pubspec.lock` is in `.gitignore`

PR #391, opened 2024-08-04
Add localizations

maelchiotti:
Add default localizations with a delegate.


Amir-P:
I can see that it's marked as a draft. Let me know when it's ready for review.

maelchiotti:
Indeed, I didn't add localizations for all the strings in the package, that's the only thing that's missing (and fixing tests). I'll tag you when I'm done.

packages/fleather/lib/src/widgets/editor_toolbar.dart
maelchiotti, line 1: @Amir-P I have an issue with the `FleatherToolbar.basic` factory. The foreground and background color buttons need to be translated, but a `BuildContext` is needed to access the `FleatherLocalizations`. Do you know what would be the best way to build these buttons in a place where a `BuildContext` is available?

Amir-P: I would say it might be fine to wrap them with `Builder` widget for now to access the context and localizations data. @maelchiotti

maelchiotti: That works indeed.

maelchiotti:
I've added the other strings, with translations in english and french. Now other languages can be supported just by adding their ARB files.

I've scrolled through all the code to find strings to translate so I hope I didn't miss any. I've also fixed the tests.

So the PR is ready for review now!

Amir-P:
Are you ok with me pushing a commit on this branch? I had too many improvements in mind that would be a bit time consuming to write and also for you to apply. @maelchiotti

maelchiotti:
Yes of course you can!

Amir-P:
Can you please take a look and let me know what you think about the changes? It also would have been better if we could completely ignore generated files from the plugin, but looks like those files don't get generated when `flutter pub get` called on the dependent project. So we have to include them. @maelchiotti

maelchiotti:
The changes look good to me!

amantoux:
do you mind if I commit a change to the repo to avoid having the **.g.dart files in the coverage stats?

amantoux:
> I've added the other strings, with translations in english and french
Cocorico :)

packages/fleather/lib/src/widgets/editor_toolbar.dart
```diff
@@ -1,12 +1,9 @@
 import 'dart:async';
 
-import 'package:fleather/src/widgets/editor.dart';
+import 'package:fleather/fleather.dart';
```
amantoux, line 3: We should avoid importing via package for internal files, can you please use relative path?

packages/fleather/lib/src/widgets/editor_toolbar.dart
```diff
@@ -1,12 +1,9 @@
...
+import 'package:fleather/l10n/l10n.dart';
```
amantoux, line 4: Same as above

packages/fleather/lib/src/widgets/link.dart
```diff
@@ -1,7 +1,8 @@
+import 'package:fleather/fleather.dart';
```
amantoux, line 1: use relative path for import?

packages/fleather/lib/src/widgets/link.dart
```diff
@@ -1,7 +1,8 @@
...
+import 'package:fleather/l10n/l10n.dart';
```
amantoux, line 2: use relative path for import?

amantoux (changes requested):
Great to see Fleather in French!
General, I think we should stick to Flutter team practice and put l10n under lib
I have a few comments on the imports

amantoux (approved):
LGTM
THanks @maelchiotti & @Amir-P

maelchiotti:
Feel free to tag me if you add a new string anytime, so I can translate it in french directly 😊

amantoux:
> Feel free to tag me if you add a new string anytime, so I can translate it in french directly 😊

It shouldn't be a problem, I'm French too 🥖

maelchiotti:
Aha awesome 🥖🥖

PR #393, opened 2024-08-05
Add clear controller feature

### LINKED ISSUE #382: Add controller's clear() method
**Is your feature request related to a problem? Please describe.**
I use FleatherEditor to send messages like in a normal chat app, so once the message is sent I would like to have an handy method to clear the whole edited message, instead of using [this workaround](https://github.com/fleather-editor/fleather/issues/265#issuecomment-1925615845)

**Describe the solution you'd like**
Instead I would like to be able to do:

`controller.clear();
`

similar to [TextEditingController.clear()](https://api.flutter.dev/flutter/widgets/TextEditingController/clear.html)

Amir-P:
Fixes #382

### DIFF
--- a/packages/fleather/lib/src/widgets/controller.dart
+++ b/packages/fleather/lib/src/widgets/controller.dart
@@ -23,7 +23,7 @@ List<String> _toggleableStyleKeys = [
 
 class FleatherController extends ChangeNotifier {
   FleatherController({ParchmentDocument? document, AutoFormats? autoFormats})
-      : document = document ?? ParchmentDocument(),
+      : _document = document ?? ParchmentDocument(),
         _history = HistoryStack.doc(document),
         _autoFormats = autoFormats ?? AutoFormats.fallback(),
         _selection = const TextSelection.collapsed(offset: 0) {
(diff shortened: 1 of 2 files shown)


packages/fleather/lib/src/widgets/controller.dart
```diff
@@ -311,6 +313,32 @@ class FleatherController extends ChangeNotifier {
       composing: TextRange.empty,
     );
   }
+
+  /// Clear the controller state.
+  ///
+  /// It creates a new empty [ParchmentDocument] and a clean edit history.
+  /// The old [document] will be closed if [closeDocument] is true.
+  ///
+  /// Calling this will notify all the listeners of this [TextEditingController]
```
amantoux, line 322: You mean `FleatherController`?

packages/fleather/test/widgets/controller_test.dart
```diff
@@ -274,6 +274,23 @@ void main() {
       // expect(controller.lastChangeSource, ChangeSource.local);
     });
 
+    test('clear', () {
```
amantoux, line 277: Shouldn't we test with `closeDocument=false`?

amantoux (changes requested):
It's look good to me overall, there are some nitpicks
Can you please take a look?

packages/fleather/test/widgets/controller_test.dart
Amir-P: Yea, I forgot about that!

amantoux (approved):
LGTM

PR #396, opened 2024-08-10
Editable elements are a repaint boundary

### LINKED ISSUE #394: The blinking cursor in the Fleather editor is causing severe heating of the phone.
I've noticed that when the editor cursor is blinking, it causes significant heating of the phone. If the editor loses focus, the phone's energy consumption returns to normal. In XCode, the increase in energy consumption is quite noticeable. I'm not sure what the cause is, as I've only used the simplest example without any additional features.

## Steps to Reproduce

Just use the simplest example.

```
import 'package:fleather/fleather.dart';
import 'package:flutter/material.dart';

void main() async {
  runApp(const MyApp());
}

class MyApp extends StatelessWidget {
  const MyApp({super.key});

  @override
    return MaterialApp(
      title: 'Flutter Demo',
      theme: ThemeData(
        colorScheme: ColorScheme.fromSeed(seedColor: Colors.deepPurple),
        useMaterial3: true,
      ),
      home: const MyHomePage(title: 'Flutter Demo Home Page'),
    );
  }
}

class MyHomePage extends StatefulWidget {
  const MyHomePage({super.key, required this.title});
  final String title;
  @override
  State<MyHomePage> createState() => _MyHomePageState();
}

class _MyHomePageState extends State<MyHomePage> {
  final controller = FleatherController(document: ParchmentDocument());
  void _incrementCounter() {
    setState(() {
    });
  }

  @override
    return Scaffold(
      appBar: AppBar(
        backgroundColor: Theme.of(context).colorScheme.inversePrimary,
        title: Text(widget.title),
      ),
      body: Column(
        children: [
          FleatherToolbar.basic(controller: controller),
          Expanded(
            child: FleatherEditor(controller: controller),
          ),
          FleatherField(controller: controller)
        ],
      ),
      floatingActionButton: FloatingActionButton(
        onPressed: _incrementCounter,
        tooltip: 'Increment',
        child: const Icon(Icons.add),
      ), // This trailing comma makes auto-formatting nicer for build methods.
    );
  }
}

```
## Environment
- OS [ iOS 17.5.1]
- Flutter version [3.22.3]
- Fleather version [1.17.0]
IMG_BCCB8547A02C-1

image

amantoux:
Closes #394

### DIFF
--- a/packages/fleather/lib/src/rendering/editor.dart
+++ b/packages/fleather/lib/src/rendering/editor.dart
@@ -861,6 +861,7 @@ class RenderEditor extends RenderEditableContainerBox
       _floatingCursorRect = null;
       _cursorController.setFloatingCursorTextPosition(null);
     }
+    markNeedsPaint();
   }
 
   void _paintFloatingCursor(PaintingContext context, Offset offset) {
--- a/packages/fleather/lib/src/rendering/editable_text_block.dart
+++ b/packages/fleather/lib/src/rendering/editable_text_block.dart
@@ -274,6 +274,9 @@ class RenderEditableTextBlock extends RenderEditableContainerBox
   // End RenderEditableBox implementation
 
   @override
+  bool isRepaintBoundary = true;
+
+  @override
   void detach() {
     _painter?.dispose();
     _painter = null;
--- a/packages/fleather/lib/src/rendering/editable_text_line.dart
+++ b/packages/fleather/lib/src/rendering/editable_text_line.dart
@@ -520,6 +520,9 @@ class RenderEditableTextLine extends RenderEditableBox {
   bool _attachedToCursorController = false;
 
   @override
+  bool isRepaintBoundary = true;
+
+  @override
   void attach(covariant PipelineOwner owner) {
     super.attach(owner);
     for (final child in _children) {
--- a/packages/fleather/lib/src/rendering/embed_proxy.dart
+++ b/packages/fleather/lib/src/rendering/embed_proxy.dart
@@ -10,6 +10,9 @@ class RenderEmbedProxy extends RenderProxyBox implements RenderContentProxyBox {
   RenderEmbedProxy({RenderBox? child}) : super(child);
 
   @override
+  bool isRepaintBoundary = true;
+
+  @override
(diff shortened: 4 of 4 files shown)


amantoux:
@Amir-P  can you please review?

packages/fleather/lib/src/rendering/editor.dart
```diff
@@ -861,6 +861,7 @@ class RenderEditor extends RenderEditableContainerBox
       _floatingCursorRect = null;
       _cursorController.setFloatingCursorTextPosition(null);
     }
+    markNeedsPaint();
```
Amir-P, line 864: I've experiment with and without this line and saw no changes. Can you explain why it was added? And maybe add a test for that so we don't forget in the future?

Amir-P (commented):
Thanks! LGTM overall. I just had a question and left you in the comments. PTAL.

Also here are the performance results against master (from the tool added in #384):

```
Analyzing tests for scrolling
Average Frame Build Time: Target: 57.08 Reference: 58.51
90th Percentile Frame Build Time: Target: 65.83 Reference: 62.51
90th Percentile Frame Build Time: Target: 68.48 Reference: 66.92
Average GPU Frame Time: Target: 0.00 Reference: 0.00
Performance tests found no regression

Analyzing tests for editing
Average Frame Build Time: Target: 430.64 Reference: 712.07
90th Percentile Frame Build Time: Target: 434.68 Reference: 715.65
90th Percentile Frame Build Time: Target: 435.07 Reference: 716.17
Average GPU Frame Time: Target: 0.00 Reference: 0.00
Performance tests found no regression
```

packages/fleather/lib/src/rendering/editor.dart
amantoux: The floating cursor stay until the next tap or edit action
In the video, notice the floating cursor that remains after releasing the floating cursor

https://github.com/user-attachments/assets/b2d6ebbe-3df2-42f0-9b60-2ede0d76eb81

Amir-P: Then I guess no need for a test. It would be really hard to test it I guess.

Amir-P (approved):
LGTM.

PR #401, opened 2024-08-11
Always keep natural order of selection boundaries

amantoux:
when setting clipboard data, selection extent can be inverted (notably on macOS) if selection was made from end to start
This fixes the issue by calling Delta::slice with natural order of selection extents


Amir-P (commented):
Nice catch. Can you please add a test to prevent regression in the future?

amantoux:
@Amir-P can you please review?

Amir-P (approved):
LGTM. Thanks.

PR #410, opened 2024-08-19
Remove generated files from VCS

amantoux:


amantoux:
Files are generated by running `flutter pub get` from the `fleather` directory and will therefore not be generated when building or getting dependencies for any app implementing _Fleather_. This would cause some extra burden for implementors

PR #415, opened 2024-08-25
Add additionalFormats to fallback and make formats public

Amir-P:
Based on #372.

Made current auto formats public for users to be able to cherrypick.

### DIFF
--- a/packages/fleather/lib/src/widgets/autoformats.dart
+++ b/packages/fleather/lib/src/widgets/autoformats.dart
@@ -28,7 +28,7 @@ class AutoFormats {
       : _autoFormats = autoFormats;
 
   /// Default set of auto formats.
-  /// 
+  ///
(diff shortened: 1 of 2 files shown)


amantoux (approved):
LGTM
I guess the natice autoformats are made public to allow users to cherrypick
It would be nice to add a comment in the PR (to avoid guessing the reason :))

amantoux (changes requested):
Maybe add test?

Amir-P:
> Maybe add test?

I found it unnecessary, but if you think it's needed I can add the test too. @amantoux

amantoux:
> > Maybe add test?
> 
> I found it unnecessary, but if you think it's needed I can add the test too. @amantoux

If not too long to write, I think it is worth

amantoux (commented):
LGTM

amantoux (approved):

PR #441, opened 2024-12-19
Update text selection when toggling checkbox

amantoux:
When toggling a checkbox the cursor is set at the offset of the checkbox's line start

Fixes issues addressed by PR #400

### DIFF
--- a/packages/fleather/lib/src/widgets/controller.dart
+++ b/packages/fleather/lib/src/widgets/controller.dart
@@ -9,7 +9,9 @@ import '../../util.dart';
 import 'autoformats.dart';
 import 'history.dart';
 
-/// List of style keys which can be toggled for insertion
+/// @docImport 'checkbox.dart';
+
+// List of style keys which can be toggled for insertion
 List<String> _toggleableStyleKeys = [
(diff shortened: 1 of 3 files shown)


amantoux:
@Amir-P  can you please review?
Essentially, this allows the check box to render it's own state while silently updating the document via the controller.
This avoids the re-build of the editor that would scroll to the cursor position while preserving the current text selection.

Amir-P (commented):
Looks like not notifying listeners is causing issues with undoing.

Steps to reproduce:

1. Create a checklist
2. Toggle it
3. Press Cmd+Z
4. Press Cmd +Z

Expected Behavior: After pressing Cmd +Z for the first time checklist should be unchecked and pressing Cmd +Z one more time reverts checklist creation.

Actual Behavior: After pressing Cmd +Z for the first time nothing happens and pressing Cmd +Z one more time reverts checklist creation.

https://github.com/user-attachments/assets/447a7a24-b009-4f84-8cb4-07a9e21982be

amantoux:
Nice catch! Will look at it

Amir-P (approved):
LGTM! Thanks!

PR #443, opened 2024-12-23
Upgrade to Flutter v3.27

Amir-P:


packages/fleather/lib/util.dart
```diff
@@ -53,3 +53,12 @@ bool isDataOnlyNewLines(Object data) {
   if (data is! String || data.isEmpty) return false;
   return RegExp('^(\n)+\$').hasMatch(data);
 }
+
+int _floatToInt8(double x) => (x * 255.0).round() & 0xff;
```
amantoux, line 57: Don't we want to wrap these function in a `value32Bits` method in an extension on `Color`?

amantoux (changes requested):
Thank Amir! I have a tiny comment PTAL

packages/fleather/lib/util.dart
Amir-P: That makes sense and reads better indeed! @amantoux

amantoux (approved):
LGTM!
Thanks!

PR #444, opened 2024-12-28
Release v1.20.0

amantoux:


packages/fleather/README.md
```diff
@@ -28,7 +28,7 @@ Add Fleather to your dependencies.
 dependencies:
   flutter:
     sdk: flutter
-  fleather: ^1.19.0
+  fleather: ^1.12.0
```
Amir-P, line 31: I think it was meant to be `1.20.0`.

Amir-P (changes requested):
Thanks, Alan. I believe we should also have a new release for Parchment as well since we had changes for Flutter 3.27. I also left a tiny comment.

amantoux:
@Amir-P Can you please review changes?

Amir-P (changes requested):
One more thing, Alan. We also need to update the parchment version in `/fleather/pubspec.yaml` file.

Amir-P (approved):
LGTM!

PR #445, opened 2025-01-01
Fix markdown decoder not recognizing all ul tokens

### LINKED ISSUE #442: Markdown encoder/decoder doesn't recognize "-" as start of unordered list item
## Steps to Reproduce

``` dart
final doc = const ParchmentMarkdownCodec().decode('''
paragraph
- a
- b
- c
''');
```

The unordered list is not recognized. 

line 39 is:

```dart
static final _ulRegExp = RegExp(r'^( *)\* +(.*)');
```
and should be:

```dart
static final _ulRegExp = RegExp(r'^( *)[-\*\+] +(.*)');
```

There is also 
```dart
    ParchmentAttribute.ul: '* ',
```
which should probably be customizable, since it depends heavily on locale and context (for example, in books - is commonly used, not *).

Here is markdown recommendation: https://www.markdownguide.org/basic-syntax/#unordered-lists. Unordered list items can start with -, *, and +.

Amir-P:
Fixes #442

### DIFF
--- a/packages/parchment/lib/src/codecs/markdown.dart
+++ b/packages/parchment/lib/src/codecs/markdown.dart
@@ -101,9 +101,9 @@ class _ParchmentMarkdownDecoder extends Converter<String, ParchmentDocument> {
       return true;
     }
 
-    if (_handleOrderedList(line, delta, style) ||
-        _handleUnorderedList(line, delta, style) ||
-        _handleCheckList(line, delta, style)) {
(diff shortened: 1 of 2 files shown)


amantoux (commented):
Thanks @Amir-P 
One question: do we want to opt-in the different characters as they seem standard?

amantoux (approved):
LGTM

PR #446, opened 2025-01-03
fix: editor crashes when multi level list uses different list types

ChizobaAmadi:
### Why 
When the list type of a multi level list is switched the editor crashes. For example, given the following list with 2 levels:
1. Ordered 1  [level 1]
   - Unordered 1 [level 2]
2. Ordered 2 [level 1]

If "1. " is changed to an unordered list, the editor crashes with a null pointer exception because it expects each list level to use the same type.

### What Changed
When looking for the next index number to apply for a numbered list check if the current level has an entry.

### Testing
Added a new multi-level numbered list test


Amir-P:
I wasn't able to reproduce the issue you've mentioned. Could you please create an issue with a sample code and steps to reprosuce the problem? A screen recording would help too. @ChizobaAmadi

amantoux:
Closing this PR since not answer @ChizobaAmadi and we could not reproduce using the editor
Feel free to reopen

PR #447, opened 2025-01-03
Fix: AutoExitBlockRule should escape indented blocks

ChizobaAmadi:
### Why
 `AutoExitBlockRule` unsets a block when the when the user enters two consecutive new line characters. This works for a single-level list but is not consistent with the behaviour of other editors for multi-level list. For example, the default IOS notes app unindents all blocks consecutively when new lines are inserted. 

### What
- Update the `AutoExitBlockRule` to check if we are in a nested level before and unindent else unset the block.

### Testing
- Added a new test case to verify the new behaviour
- Tested by running the example app


Amir-P (commented):
Hey @ChizobaAmadi , Thanks for contributing to Fleather!

Looks like you've based this branch on the `fix/multi_level_different_list_type` and brought those changes here too. Can you please fix that so that we can only review the changes regarding `AutoExitBlockRule`?

ChizobaAmadi:
Closing this PR and opening a new [one](https://github.com/fleather-editor/fleather/pull/447)

PR #450, opened 2025-01-22
Fix: AutoExitBlockRule should escape indented blocks

ChizobaAmadi:
### Why
AutoExitBlockRule unsets a block when the when the user enters two consecutive new line characters. This works for a single-level list but is not consistent with the behaviour of other editors for multi-level list. For example, the default IOS notes app unindents all blocks consecutively when new lines are inserted.

### What
Update the AutoExitBlockRule to check if we are in a nested level before and unindent else unset the block.

### Testing
Added a new test case to verify the new behaviour
Tested by running the example app


Amir-P (approved):
Thanks! LGTM overall. We just need to update the documentation for `AutoExitBlockRule` to mention the unindenting. Could you update that too? @ChizobaAmadi 

@amantoux Could you also take a look at the PR?

amantoux:
> Thanks! LGTM overall. We just need to update the documentation for `AutoExitBlockRule` to mention the unindenting. Could you update that too? @ChizobaAmadi
> 
> @amantoux Could you also take a look at the PR?

LGTM too. Thanks @ChizobaAmadi

amantoux:
also @ChizobaAmadi can you run fix the code format issues please?

PR #451, opened 2025-01-22
[Draft] support apply link directly to text

ChizobaAmadi:
It would be great if a user could directly paste a link to a text. The stock IOS notes app detects if the clipboard data is a link and then applies it to the current text selection, if it is not a link it replaces the selected text (same as Fleather).

I'd like to know if my current approach is the right approach! It seems to work correctly on mobile.


Amir-P:
Hey @ChizobaAmadi, is it still a draft?

amantoux:
Closing as stale

PR #454, opened 2025-01-31
Fix null check on null value error in _onChangedClipboardStatus

### LINKED ISSUE #435: RawEditorState._onChangedClipboardStatus
截屏2024-10-23 20 24 03

I received such exceptions through Firebase; there might be an oversight in the Null check operator.

Amir-P:
Fixes #435


packages/fleather/example/ios/Podfile.lock
amantoux, line 1: Do we want to put this in this PR?

amantoux (commented):
LGTM, one small question though, can you please take a look?

packages/fleather/example/ios/Podfile.lock
Amir-P: Maybe not. If you find it inappropriate I can exclude the change. @amantoux

amantoux: Nevermind

PR #470, opened 2025-03-20
Upgrade Gradle and use plugins in example app

Amir-P:


amantoux (approved):
Not a specialist on Android but LGTM

PR #473, opened 2025-03-27
Add ko Localization

ezzooooo:
i'm korean developer.
i will use fleather, but it can't support 'ko' locale.

so, i have searched common text editors. and then i made "ko.arb".
if you confirm, and please merge.

thank you.


amantoux:
Thank you @ezzooooo
LGTM!

amantoux (approved):

amantoux:
@Amir-P can you take a look?

PR #476, opened 2025-04-06
Add pt_BR localization

marcio-ota:


Amir-P (changes requested):
Thanks for contributing!
The translations look good to me. But same as #477, you need to run `flutter gen-l10n` to generate the dart classes for your translations.

PR #477, opened 2025-04-06
Add pt localization

marcio-ota:


Amir-P (changes requested):
Thanks for contributing!
The translations look good to me. But you need to run `flutter gen-l10n` to generate the dart classes for your translations.

PR #478, opened 2025-04-06
add pt, pt_BR localization

marcio-ota:


Amir-P (approved):
LGTM. Thanks again!
@amantoux Do you want to take a look too?

amantoux (approved):
All good here

PR #480, opened 2025-04-08
Release v1.21.0

amantoux:


Amir-P (changes requested):
Thanks, Alan. Don't we also need a release for parchment as we had a change and removed a deprecated analysis rule?

amantoux:
> Thanks, Alan. Don't we also need a release for parchment as we had a change and removed a deprecated analysis rule?

@Amir-P How does this affect the users? I'm not sure

PR #487, opened 2025-04-26
Handle multi-level list followed by another list or a plain paragraph

### LINKED ISSUE #459: List Conversion Issues in HTML Output
## Description

When converting content that contains nested ordered and unordered lists from Fleather to HTML, the resulting HTML does not correctly represent the intended list structure. This issue appears under two conditions:

- When there is a newline between the two lists.

- When there is no newline between the two lists.

## Steps to Reproduce

1. Create an Ordered List

- Item 1: Level 1 - “1”
- Item 2: Level 1 - “2”
- Item 3: Level 2 - “1”

2. Create an Unordered List (Placed Below the Ordered List)

- Item 1: Level 1 - “1”
- Item 2: Level 2 - “1”

3. Convert the Document to HTML with ParchmentHtmlCodec.encode()

## Environment
OS: macOS 15.2
Flutter 3.27.2
Fleather 1.19.0

### In The Editor:

Image

### HTML Output:

Image

### In The Editor:

Image

### HTML Output:

Image

It appears that there is a problem with how Fleather encodes nested lists/lists that are below each other into HTML. The bug differs based on whether a newline exists between the ordered and unordered lists, leading to inconsistent HTML output.

Any insights or fixes to ensure consistent HTML conversion for both scenarios would be greatly appreciated.

amantoux:
Fixes #459

### DIFF
--- a/packages/parchment/lib/src/codecs/html.dart
+++ b/packages/parchment/lib/src/codecs/html.dart
@@ -176,6 +176,36 @@ class _ParchmentHtmlEncoder extends Converter<ParchmentDocument, String> {
     return currentLevel < candidateLevel;
   }
 
+  // Check if both attributes are lists of different type with same indentation
+  bool isDifferentListTypeWithSameIndentationLevel(
+      ParchmentStyle parent, ParchmentStyle child) {
+    final currentListAttribute = parent.values.firstWhereOrNull(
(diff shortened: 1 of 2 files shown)


Amir-P (approved):
LGTM!

PR #491, opened 2025-04-28
Fix floating cursor position issue by fixing getLocalRectForCaret

### LINKED ISSUE #483: Incorrect floating cursor behavior
When activating floating cursor with `FleatherEditor.padding.start/left` greater than 0.0, the floating cursor is not correctly positioned

## Steps to Reproduce

1. In example (padding start = 16), activate floating cursor
2. Release it
3. Start again
4. Notice how the cursor changed position without performing any movement with keyboard

https://github.com/user-attachments/assets/cef56a1b-1b34-46cc-95e4-33bb8dab1c5f

## Environment
- iOS
- Flutter 3.29
- Fleather v1.21

Amir-P:
This is an attempt to fix #483 while preventing issue caused by #484. Since we are already adding padding offset to children, we shouldn't change the paint offset based on padding.

Before (Example app):
image
After:
image

### DIFF
--- a/packages/fleather/lib/src/rendering/editor.dart
+++ b/packages/fleather/lib/src/rendering/editor.dart
@@ -187,17 +187,7 @@ class RenderEditor extends RenderEditableContainerBox
     markNeedsSemanticsUpdate();
   }
 
-  Offset get paintOffset {
-    final double dx;
-    if (padding is EdgeInsets) {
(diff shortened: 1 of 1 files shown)


amantoux (approved):
Thanks @Amir-P ! This is a better solution to address the problem

Issue #513, opened 2025-06-12
Read-only mode iOS context menu exception


ibrahimdevs:
Firstly, thanks for your great effort and made this editor real. 

## Steps to Reproduce

1. Set editor readonly mode: readOnly: true
2. Select some text to display context menu
3. Problem: Throws error, because the latest version is using iOS native context menu which needs active text input connection. But readonly mode I think there is not.

There is a quick solution in my mind but I'm not sure it is the best solution.

```
/// Default implementation of a widget builder function for context menu.
Widget defaultContextMenuBuilder(BuildContext context, EditorState editorState) {
  if (defaultTargetPlatform == TargetPlatform.iOS &&
      SystemContextMenu.isSupported(context) && editorState.widget.readOnly == false) {
    return SystemContextMenu.editor(editorState: editorState);
  }
  return AdaptiveTextSelectionToolbar.buttonItems(
    buttonItems: editorState.contextMenuButtonItems,
    anchors: editorState.contextMenuAnchors,
  );
}
```

## Environment
- OS = iOS
- Flutter version = 3.32.2
- Fleather version = 1.23.0

## Logs

```
======== Exception caught by widgets library =======================================================
The following assertion was thrown building SystemContextMenu(dirty, dependencies: [MediaQuery, _LocalizationsScope-[GlobalKey#a6a51]], state: _SystemContextMenuState#856d6):
Currently, the system context menu can only be shown for an active text input connection
'package:flutter/src/services/text_input.dart':
Failed assertion: line 2734 pos 7: 'TextInput._instance._currentConnection != null'

The relevant error-causing widget was: 
  SystemContextMenu SystemContextMenu:file:///Users/ibrahim/.pub-cache/hosted/pub.dev/fleather-1.23.0/lib/src/widgets/editor.dart:53:30
When the exception was thrown, this was the stack: 
#2      SystemContextMenuController.showWithItems (package:flutter/src/services/text_input.dart:2734:7)
#3      _SystemContextMenuState.build (package:fleather/src/widgets/system_context_menu.dart:118:36)
#4      StatefulElement.build (package:flutter/src/widgets/framework.dart:5823:27)
#5      ComponentElement.performRebuild (package:flutter/src/widgets/framework.dart:5715:15)
#6      StatefulElement.performRebuild (package:flutter/src/widgets/framework.dart:5874:11)
#7      Element.rebuild (package:flutter/src/widgets/framework.dart:5427:7)
[...]

PR #515, opened 2025-06-13
Not display iOS native context menu on read only mode.

ibrahimdevs:
iOS native context menu needs active text input connection. If the editor is on read-only mode, there is no active text input connection and Flutter throws exception.

Fix: not show iOS native context menu if the editor is on read-only mode.

Solves: https://github.com/fleather-editor/fleather/issues/513


Amir-P (commented):
Thanks for contributing @ibrahimdevs . Can you please add a test for the read only case on iOS?

Amir-P (approved):
LGTM. Thanks again!
Do you wanna take a look as well? @amantoux

amantoux (approved):
LGTM
Thank you for the contribution & the warm support

PR #528, opened 2025-07-20
Re-initialized ordered list index when no current index available

### LINKED ISSUE #527: Editing the following specific format will result in an error
Editing the following specific format will result in an error

Image

Image

amantoux:
Fixes #527


Amir-P (approved):
LGTM. Can you please create an issue to fix the behavior as this only prevents the exception but the underlying issue is still there. @amantoux

PR #530, opened 2025-08-01
Allow the editor in read-only to tighten to the actual necessary width

amantoux:
This PR allows editor to shrink to max content size in read-only

To acheive this, one can use a new paramter `textWidthBasis`
```dart
FleatherEditor(
   controller: controller,
   scrollable: false,
   readOnly: true,
   enableInteractiveSelection: false,
   textWidthBasis: TextWidthBasis.longestLine,
);
```

### Before
image

### After
image

### DIFF
--- a/packages/fleather/lib/src/widgets/field.dart
+++ b/packages/fleather/lib/src/widgets/field.dart
@@ -91,6 +91,12 @@ class FleatherField extends StatefulWidget {
   /// the text field from the clipboard.
   final bool enableInteractiveSelection;
 
+  /// Defines how to measure the width of the rendered when [readOnly] is `true`
+  /// text. Otherwise the value is ignored and forced to [TextWidthBasis.parent]
+  ///
+  /// Default to [TextWidthBasis.parent].
+  final TextWidthBasis textWidthBasis;
+
   /// The minimum height to be occupied by this editor.
   ///
   /// This only has effect if [scrollable] is set to `true` and [expands] is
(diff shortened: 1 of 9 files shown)


packages/fleather/lib/src/widgets/text_line.dart
```diff
@@ -136,6 +137,7 @@ class _TextLineState extends State<TextLine> {
         text: text,
         textAlign: textAlign,
         strutStyle: strutStyle,
+        textWidthBasis: editorState.widget.textWidthBasis,
```
Amir-P, line 140: I think it would be better to provide the `textWidthBasis` as a constructor parameter, rather than the editor state.

packages/fleather/lib/src/widgets/field.dart
```diff
@@ -91,6 +91,12 @@ class FleatherField extends StatefulWidget {
   /// the text field from the clipboard.
   final bool enableInteractiveSelection;
 
+  /// Defines how to measure the width of the rendered when [readOnly] is `true`
```
Amir-P, line 94: I think it was meant to be "Defines how to measure the width of the rendered text when [readOnly] is true. ...", wasn't it?

Amir-P (commented):
We really needed it! Thanks! I just have two small comments. PTAL.

amantoux:
@Amir-P Fixed accordingly, can you please take a look?

PR #583, opened 2026-05-21
Support Flutter 3.44 text input changes

john-unger:
Adds compatibility for Flutter 3.44 text input API changes, updates deprecated text input style handling, and fixes the pt_BR ARB filename so localization generation succeeds.


Amir-P (commented):
Hi @john-unger, thanks for contributing! I think you also need to update the .github/workflows/fleather.yml to use the latest Flutter version for running the workflow. Other than, it looks good to me.

john-unger:
thank you @Amir-P i have made that update, and fixed the locale issue on the GHA with pt-BR. Can you take it from here?

packages/fleather/lib/src/widgets/editor_input_client_mixin.dart
```diff
@@ -200,6 +200,9 @@ mixin RawEditorStateTextInputClientMixin on EditorState
     // no-op
   }
```
amantoux, line 202: Flutter's `EditableTextState` implements this method with
```dart
@override
bool onFocusReceived() {
  if (mounted && !_hasFocus && widget.focusNode.canRequestFocus) {
    widget.focusNode.requestFocus();
    return true;
  }
  return false;
}
```

Should we implement something similar?

amantoux (commented):
I have a small comment, can you please check?

amantoux (approved):
LGTM, thanks

PR #586, opened 2026-07-24
feat(fleather): add iPad Scribble support

zhpoo:
Changes:

- expose `stylusHandwritingEnabled` from `FleatherEditor`, `FleatherField`, and `RawEditor`;
- implement Flutter's `ScribbleClient` lifecycle for editable iOS editors;
- report UTF-16 character, caret, composing, size, and transform geometry to the text input connection;
- render and remove Scribble placeholders without changing the Parchment document;
- support selection boxes across Fleather block nodes;
- accept Scribble handwriting across the full editor surface without moving the current selection on pen-down;
- connect Fleather history to the iOS system undo manager for Pencil toolbar undo and redo;
- reject Scribble input whose center is outside the editor bounds;
- keep Scribble disabled at runtime on non-iOS platforms and in read-only editors.

### DIFF
--- a/packages/fleather/lib/src/widgets/editor.dart
+++ b/packages/fleather/lib/src/widgets/editor.dart
@@ -183,6 +183,11 @@ class FleatherEditor extends StatefulWidget {
   /// the text field from the clipboard.
   final bool enableInteractiveSelection;
 
+  /// Whether this editor supports platform stylus handwriting.
+  ///
+  /// On iPadOS this enables Apple Pencil Scribble.
+  final bool stylusHandwritingEnabled;
+
   /// Defines how to measure the width of the rendered text when [readOnly] is
   /// `true`. Otherwise the value is ignored and forced to
   /// [TextWidthBasis.parent]
@@ -317,6 +322,8 @@ class FleatherEditor extends StatefulWidget {
       this.autocorrect = true,
       this.enableSuggestions = true,
       this.enableInteractiveSelection = true,
+      this.stylusHandwritingEnabled =
+          EditableText.defaultStylusHandwritingEnabled,
       this.textWidthBasis = TextWidthBasis.parent,
       this.minHeight,
       this.maxHeight,
@@ -341,11 +348,15 @@ class FleatherEditor extends StatefulWidget {
 class _FleatherEditorState extends State<FleatherEditor>
     implements EditorTextSelectionGestureDetectorBuilderDelegate {
   GlobalKey<EditorState>? _editorKey;
+  FocusNode? _internalFocusNode;
 
   bool _showSelectionHandles = false;
 
   @override
   GlobalKey<EditorState> get editableTextKey => widget.editorKey ?? _editorKey!;
+
+  FocusNode get effectiveFocusNode =>
+      widget.focusNode ?? (_internalFocusNode ??= FocusNode());
 
   @override
   bool get forcePressEnabled => true;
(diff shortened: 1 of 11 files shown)


packages/fleather/lib/src/widgets/history.dart
```diff
@@ -15,40 +19,123 @@ import '../util.dart';
...
+        UndoManager.client = null;
+      }
+      return;
+    }
+    UndoManager.client = this;
+    _updateUndoState();
+  }
+
+  void _updateUndoState() {
+    if (defaultTargetPlatform == TargetPlatform.iOS &&
```
Amir-P, line 102: I believe we should skip the check for iOS platform here. If we are the client, we should update the state anyway.

packages/fleather/lib/src/widgets/history.dart
```diff
@@ -15,40 +19,123 @@ import '../util.dart';
...
   final FleatherController controller;
 
   @override
   State<FleatherHistory> createState() => _FleatherHistoryState();
 }
 
-class _FleatherHistoryState extends State<FleatherHistory> {
+class _FleatherHistoryState extends State<FleatherHistory>
+    with UndoManagerClient {
+  Timer? _historyStateTimer;
```
Amir-P, line 44: I don't see a reason for introducing a timer and periodically updating the undo state as the undo state is retrieved from the controller and it would notify the listeners if the state changes. Am I missing something?

packages/fleather/lib/src/widgets/editor_input_client_mixin.dart
```diff
@@ -366,6 +366,28 @@ mixin RawEditorStateTextInputClientMixin on EditorState
...
+    required Rect caretRect,
+    required Rect composingRect,
+  }) {
+    if (!hasConnection) {
+      return;
+    }
+    final size = Size(
+        min(renderEditor.size.width,
+                renderEditor.maxContentWidth ?? double.infinity) -
+            renderEditor.padding.horizontal,
```
Amir-P, line 381: I think we should just use the `renderEditor.size` here and not take padding or content width into account as user should be able to write with stylus anywhere within the editor area. This is also what Flutter's `EditableText` does:
```dart
  void _updateSizeAndTransform() {
    final Size size = renderEditable.size;
    final Matrix4 transform = renderEditable.getTransformTo(null);
    _textInputConnection!.setEditableSizeAndTransform(size, transform);
  }
```

packages/fleather/lib/src/widgets/editor.dart
```diff
@@ -341,12 +348,16 @@ class FleatherEditor extends StatefulWidget {
...
     implements EditorTextSelectionGestureDetectorBuilderDelegate {
   GlobalKey<EditorState>? _editorKey;
+  FocusNode? _internalFocusNode;
 
   bool _showSelectionHandles = false;
 
   @override
   GlobalKey<EditorState> get editableTextKey => widget.editorKey ?? _editorKey!;
 
+  FocusNode get effectiveFocusNode =>
```
Amir-P, line 358: I think it's better to make the focus node required and non-optional for the `RawEditor` now that we have lifted the internal focus node creation up. It's gonna be a breaking change but I don't think many would be affected. WDYT? @amantoux

Amir-P (changes requested):
Thanks for working on this! I took a look at the changes and left some comments. Please take a look. @zhpoo 

I don't own an iPad so I'm not gonna be able to test and verify the behavior. Will you be able to do that? @amantoux

PR #591, opened 2026-09-11
Selector toolbar position

amantoux:
Color and heading selectors can cover the formatting toolbar when it sits above the keyboard. Position selectors below the presenter when there is room, and flip above it when that side offers more space. Use the full toolbar's vertical bounds while keeping horizontal alignment with the button, and preserve keyboard and safe-area constraints.

### DIFF
--- a/packages/fleather/lib/src/widgets/editor_toolbar.dart
+++ b/packages/fleather/lib/src/widgets/editor_toolbar.dart
@@ -1274,14 +1274,32 @@ class SelectorScopeState extends State<SelectorScope> {
     final RenderBox presenter = context.findRenderObject() as RenderBox;
     final RenderBox overlayBox =
         overlay.context.findRenderObject() as RenderBox;
-    final offset = Offset(0.0, presenter.size.height);
+    final presenterRect = Rect.fromPoints(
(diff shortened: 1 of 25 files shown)


packages/fleather/lib/src/widgets/editor_toolbar.dart
```diff
@@ -1395,13 +1411,29 @@ class _SelectorLayout extends SingleChildLayoutDelegate {
...
+        _selectorScreenPadding -
+        subScreen.top -
+        padding.top -
+        _selectorScreenPadding;
+    // Prefer below the presenter, but flip above it when that offers more room.
+    final bool showAbove =
+        childSize.height > spaceBelow && spaceAbove > spaceBelow;
+    final double y = showAbove
+        ? presenter.top - _selectorScreenPadding - childSize.height
+        : presenter.bottom + _selectorScreenPadding;
```
Amir-P, line 1435: Thanks for working on this and it works good except I believe there is extra padding now as you can see from the screenshots. I believe you shouldn't account for `_selectorScreenPadding` here as they would be added in `_fitInsideScreen` if needed (to push it away from screen edges).

Before:
Image
Image

After:
Image
Image

